PR #4, opened 2022-02-11
Add lexical-binding and options to startup quicker

mfxarc:


japhir:
Hey this looks cool, what does the lexical-bindings local-variable do, precisely? I was trying to read the [manual on it](https://www.gnu.org/software/emacs/manual/html_node/elisp/Lexical-Binding.html) and understood from that that it's forcing the variable lookup to only look at local variables within the file. Is that correct? I'm new to writing elisp and would like an ELI5 if you have the time :).

daviwil:
This is great, thank you very much!

daviwil:
@japhir Lexical binding is a different way to achieve variable lookup when running the code which removes dynamic variable resolution for local variables and replaces it with a much faster lookup based that is generated by the compiler.  Since it changes the semantics of how Emacs Lisp code runs, it has to be turned on explicitly for every `.el` file.

PR #9, opened 2022-02-12
load vertico directory from extensions folder

ptillemans:
When requiring the **rational-completion** I got an error that **vertico-directory** could not be required.

When checking the vertico package I noticed it was in an *extensions* subfolder which is not added to the load-path so I added the filename parameter on the require form.

### DIFF
--- a/modules/rational-completion.el
+++ b/modules/rational-completion.el
@@ -18,7 +18,7 @@ folder, otherwise delete a word"
     (backward-kill-word arg)))
 
 (require 'vertico)
-(require 'vertico-directory)
+(require 'vertico-directory "extensions/vertico-directory.el")
 
 (with-eval-after-load 'evil
   (define-key vertico-map (kbd "C-j") 'vertico-next)


JulienTant:
Can confirm, I ran into the same problem and the proposed solution is working 👍🏻

daviwil:
Excellent, thanks a lot @ptillemans! I actually didn't know you could do this :)

jeffbowman:
This solution loads the file from source, which loses the advantage of loading the byte compiled file. Would it be better to just add the extensions folder to the load path so the require could pick either the el or elc file to load? For example: `(add-to-list 'load-path (expand-file-name "straight/build/vertico/extensions" straight-base-dir))` accomplishes the same thing as the solution here with the advantage the require could pick the elc file.

daviwil:
@jeffbowman we could definitely give that a shot.  It's a little unfortunate if `require` doesn't do the right thing automatically.

jeffbowman:
`require` does "the right thing" when you don't mention the file... thus `(require 'vertico-directory)` will first try to load the .elc file if it exists and then the .el file if not. However, `(require 'vertico-directory "extensions/vertico-directory.el")` assumes you know what you are doing and loads that file regardless if the .elc exists.

daviwil:
Gotcha, makes sense.

PR #12, opened 2022-02-12
defaults, editing, ui tweaks

japhir:


modules/rational-ui.el
```diff
@@ -22,11 +22,24 @@ the same key names as accepted by `set-face-attribute'")
       doom-modeline-buffer-file-name-style 'truncate-except-project)
 
 ;; Make `describe-*' screens more helpful!
-(require 'helpful)
+(straight-use-package 'helpful)
```
japhir, line 25: I don't think helpful is built-in, but I also don't use straight-use-package yet. Is this correct?

modules/rational-ui.el
```diff
@@ -22,11 +22,24 @@ the same key names as accepted by `set-face-attribute'")
...
 (global-set-key [remap describe-function] #'helpful-callable)
 (global-set-key [remap describe-variable] #'helpful-variable)
 (global-set-key [remap describe-key] #'helpful-key)
 (global-set-key (kbd "C-h F") #'helpful-function)
 (global-set-key (kbd "C-h C") #'helpful-command)
 
+;; also add some examples
+(straight-use-package 'elisp-demos)
```
japhir, line 33: this adds the example function calls that I mentioned during the stream

japhir:
Thanks @JulienTant, sounds good :). I've removed the `(prefer-coding-system 'utf-8-unix)` line and moved the remainder to init.el.

modules/rational-ui.el
daviwil: I have it higher up in this file. The strategy I've been using so far is to put all the package ensure expressions at the top so its clearer which packages get installed by a module.

japhir:
Should be ready to merge if all lines in this are desired now :)

daviwil:
Thanks a bunch!

PR #13, opened 2022-02-12
(rational-default) Turn on recentf mode in rational-defaults

JulienTant:
recentf mode is pure awesomeness and I really struggle to understand why it's not turned on by default in emacs.

### DIFF
--- a/init.el
+++ b/init.el
@@ -41,6 +41,16 @@ straight.el or Guix depending on the value of
 (defvar rational-config-file (expand-file-name "config.el" rational-config-path)
   "The user's configuration file.")
 
+;; Defines the user configuration var and etc folders
+;; and ensure they exists.
+(defvar rational-config-etc-directory (expand-file-name "etc/" rational-config-path)
+  "The user's configuration etc/ folder.")
+(defvar rational-config-var-directory (expand-file-name "var/" rational-config-path)
+  "The user's configuration var/ folder.")
+
+(mkdir rational-config-etc-directory t)
+(mkdir rational-config-var-directory t)
+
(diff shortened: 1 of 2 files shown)


daviwil (commented):
Thanks! Should we also add a limit to the number of recent items? An unbounded number of items could make startup time slower, but maybe there's already a default set?

jeffbowman:
Yes! I second this. Also great idea to use the `rational-emacs/` folder to store recent files, thus not cluttering version control or even needing to provide that directory in gitignore.

daviwil:
@JulienTant Yep, great idea.  I think it'd be even better if we could make it a subfolder under `rational-emacs` which can be added to `.gitignore` so that it doesn't cause issues with Git when we add the ability to pull updates to the base config.

JulienTant:
Hey, if we put it in the user rational-emacs/ folder, no need to gitignore anything it will be out of this repository.

daviwil:
Whoops, I misunderstood.  That sounds fine!

daviwil:
However, maybe it's still a good idea to use a subfolder that makes it clear that this is not a set of files that need to be checked into one's own dotfiles, sort of like what `no-littering` does (which we could consider using, even.

JulienTant:
Until `no-littering` is introduced, I "copied" the general concept and moved the `recentf` file in the `{USER-CONFIG}/var` folder.
Once/if `no-littering` is introduced, feel free to ping me to refactor this part using it.

daviwil:
Thanks a bunch for doing that! I think this is a fine approach.

PR #14, opened 2022-02-13
Add org and files-management modules

mfxarc:
- Add a simple org module with some sane defaults
- Add a file-management module to serve as a good environment to manage files. 

  - IDEA: Add open-with or another package to handle externally opening files
  - IDEA: Prefer using built-in tools to handle files (e.g. doc-view, image-mode...)

### DIFF
--- a/modules/rational-files-management.el
+++ b/modules/rational-files-management.el
@@ -1,0 +1,19 @@
+;;; rational-files-management.el -*- lexical-binding: t; -*-
+
+(require 'dired)
+
+;; Always delete and copy recursively
+(setq dired-recursive-deletes 'always)
+(setq dired-recursive-copies 'always)
+
+;; Create destinations directories if needed
+(setq dired-create-destination-dirs 'always)
+
+;; Refresh buffers automatically
+(add-hook 'dired-mode-hook 'auto-revert-mode)
+
+;; Display directories first
+(setq dired-listing-switches "-vAlahF --group-directories-first")
+
+(provide 'rational-files-management)
+;;; rational-files-management ends here
--- a/modules/rational-org.el
+++ b/modules/rational-org.el
@@ -1,0 +1,18 @@
+;;; rational-org.el  -*- lexical-binding: t; -*-
+
+(straight-use-package 'org)
+(straight-use-package 'org-appear)
+
+;; Return or left-click with mouse follows link
+(setq org-return-follows-link t)
+(setq org-mouse-1-follows-link t)
+
+;; Display links as the description provided
+(setq org-descriptive-links t)
+
+;; Hide markup markers
+(setq org-hide-emphasis-markers t)
+(add-hook 'org-mode 'org-appear-mode)
+
+(provide 'rational-org)
+;;; rational-org.el ends here


modules/rational-files-management.el
```diff
@@ -0,0 +1,19 @@
+;;; rational-files-management.el -*- lexical-binding: t; -*-
```
daviwil, line 1: I could call this one `rational-file-management`

modules/rational-files-management.el
```diff
@@ -0,0 +1,19 @@
...
+(setq dired-recursive-copies 'always)
+
+;; Create destinations directories if needed
+(setq dired-create-destination-dirs 'always)
+
+;; Refresh buffers automatically
+(add-hook 'dired-mode-hook 'auto-revert-mode)
+
+;; Display directories first
+(setq dired-listing-switches "-vAlahF --group-directories-first")
```
daviwil, line 16: We might need to add a check for whether we're on Windows and use a different set of params, but I'm not sure what Dired actually uses for dir listings on Windows...

daviwil:
This is a great start, thanks a lot!  I've been annoyed a bit by `openwith-mode` (sometimes emojis appearing in ERC chats will make an image viewer open), so I wonder if we can find an option for external file loads that works only inside of Dired without hijacking `find-file`.

mfxarc (commented):
For now, we could use a checker to see if the user is using Linux
```
(defconst ON-LINUX (eq system-type 'gnu/linux))
(when ON-LINUX
  (setq dired-listing-switches "-vlah --group-directories-first"))
```

modules/rational-files-management.el
gan-ainm: The -A option suppresses listing the . and .. entries while -a re-introduces them again, so it should be either -A or -a, not both options at the same time.

init.el
```diff
@@ -37,6 +37,12 @@ straight.el or Guix depending on the value of
         (message "Package '%s' does not appear to be installed by Guix!"))
     (straight-use-package package)))
 
+;; Check the system used
+(defconst ON-LINUX   (eq system-type 'gnu/linux))
```
daviwil, line 41: Since the type of OS can only ever be one of these options at a time, perhaps we should boil it down to a single variable?  Maybe something like `rational-current-operating-system` which has the values of `windows`, `macos`, or `linux`?

modules/rational-org.el
```diff
@@ -0,0 +1,18 @@
+;;; rational-org.el  -*- lexical-binding: t; -*-
+
+(straight-use-package 'org)
```
daviwil, line 3: Great starting point!

abarocio80: I wonder if it will not be better to use the default `org` package that comes with Emacs? If any other package requires a newer version, then it will be updated (as with `org-roam`).

init.el
ajxn: In my configuration I have all those operating systems, but I also check for graphical system, like this.

      (memq window-system '(mac ns x))

And I uses `darwin` to set my keyboard for my Mac, that is just OS, no GUI related.

modules/rational-org.el
abarocio80: I mean: it would be "rational" to use what ever emacs comes with, don't you think? Is this not the general approach of this project?

daviwil (approved):
I'll go ahead and merge this and make a few tweaks.  Thanks a lot!

PR #18, opened 2022-02-14
Jackson

ajxn:
This is several different small pulls that could be useful.

* Max line length are usually set to about under 80 characters.
* Org-mode want to use GNU elpa repositories, so they should probably be added first.
* There are already a function for calculation the execution time to init Emacs.
* Customizing settings moved to a file of it's own and load those when starting (maybe in early-init.el).
* Added non nil to NOERROR argument of load to not bail out if the file doesn't exist (or error loading), then no need for check if file exists.
* Just let git ignore standard Emacs temp files.

But as I am new to using GitHub, I guess I have to figure this out how to do this in Emacs magit.

### DIFF
--- a/modules/rational-use-package.el
+++ b/modules/rational-use-package.el
@@ -4,8 +4,15 @@
 
 (require 'package)
 
-(setq package-archives '(("melpa" . "https://melpa.org/packages/")
-                         ("elpa" . "https://elpa.gnu.org/packages/")))
+;; gnu and nongnu is needed for packages like Org-mode
+(setq package-archives
+      '(("gnu" . "https://elpa.gnu.org/packages/")))
+(add-to-list package-archives
+	     '("nongnu" . "https://elpa.nongnu.org/nongnu") 'append)
+(add-to-list package-archives
+	     '("melpa" . "https://melpa.org/packages/") 'append)
+(add-to-list package-archives
+	     '("elpa" . "https://elpa.gnu.org/packages/") 'append)
 
 (package-initialize)
 (unless package-archive-contents
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ straight/
 transient/
 eshell/
 projects
+*~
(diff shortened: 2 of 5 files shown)


ajxn:
(Need to learn to name branches properly, for one :-)

modules/rational-use-package.el
```diff
@@ -4,8 +4,15 @@
 
 (require 'package)
 
-(setq package-archives '(("melpa" . "https://melpa.org/packages/")
-                         ("elpa" . "https://elpa.gnu.org/packages/")))
+;; gnu and nongnu is needed for packages like Org-mode
+(setq package-archives
```
daviwil, line 8: I'd update this to:

```emacs-lisp
(setq package-archives '(("gnu" . "https://elpa.gnu.org/packages/")
                         ("nongnu" . "https://elpa.nongnu.org/nongnu/")
                         ("melpa" . "https://melpa.org/packages/")))
```

daviwil (commented):
Thanks Anders!  Only one thing I'd tweak here, the rest of the changes are great!

modules/rational-use-package.el
jeffbowman: Another thought, the "gnu" repo is default in 27.x, while the "gnu" and "nongnu" are default in 28.x, so, only need to add "nongnu" if emacs version is 27.x and then always add "melpa".  With that in mind, the `add-to-list` approach allows adding a check on emacs version around the "nongnu" repo. Thus:

    (when (< (string-to-number emacs-version) 28)
      (add-to-list 'package-archives (cons "nongnu" "https://elpa.nongnu.org/nongnu/")) 'append)
    (add-to-list 'package-archives (cons "melpa" "https://melpa.org/packages/") 'append)

JulienTant:
I'm having a hard time making sense of this `80 characters` in a `text file`, it just makes the text more difficult to read.

PR #24, opened 2022-02-15
Prefer `customize-set-variable' for customizing packages

jeffbowman:


japhir:
I wondered why this would be preferred, and came across [this nice stackoverflow discussion](https://emacs.stackexchange.com/questions/102/advantages-of-setting-variables-with-setq-instead-of-custom-el).

daviwil:
Yep, this is definitely the right approach, it's just a little annoying to have to call such a long function name to set variables.  Let's do the right thing here and use it, thanks @jeffbowman!

PR #26, opened 2022-02-15
Conditionally add elpas to package archive based on Emacs version

jeffbowman:
Based on discussion from #18 provide the suggestion for only adding elpas based on Emacs version.


daviwil:
Looks great, thanks a lot!

Issue #33, opened 2022-02-15
set (relative) line-numbers to on by default or in one of the modules(defaults or ui maybe)


erikLundstedt:

jeffbowman:
I agree, turning on line numbers does make sense as a default option. I prefer to defer to the eventual user to make them relative if that is their preference. That being said, perhaps a defun could be provided that would configure it that way so they only have to call it?

JulienTant:
As long as we have to way to turn if off if desired :P I usually prefer refering to my modline to get my position and have as much "screen space" as possible for the content of my page.

Note: I'm having eye issues and let's say I "zoom" a lot :P
![image](https://user-images.githubusercontent.com/785518/154172523-9d06cf4f-6244-4cbf-86d3-2a6192aadceb.png)

japhir:
I still don't understand what problem showing line numbers solves: I guess with relative lines you can jump up or down quickly week 10j, but at that point I'd prefer to jump via search

erikLundstedt:
obviously we should be able to turn it off(although i personally use it globaly)
i just feel its a really sensible default to have it on

jeffbowman:
@japhir it solves the same problem as seeing the line number in the modeline, however, more of them may be seen at a glance than just the one where the point is. 

I made some comments on another issue that might be relevant here: https://github.com/SystemCrafters/rational-emacs/issues/35#issuecomment-1042018008

ajxn:
I would suggest that use a `customize-set-variable` for this.  And that it is turned on in `prog-mode`, as I can't see the use of it in any `text-mode`.

Alternatively copy some example code into `example-config.el` where the user can make a decition of their own, but they see how to turn it on/off there.  As it is now, there are nothing in `example-config.el` and that is not much of an example.  :-)
But then again, there are some examples in `README.org`.

jeffbowman:
Can this be closed? The feature is implemented in `rational-ui.el`, is that sufficient for this issue?

jeffbowman:
Closing this issue since there has been no additional discussion and #99 is also closed.

PR #40, opened 2022-02-17
Add line-numbering configuration

angrybacon:


erikLundstedt:
this seems good


this should close
https://github.com/SystemCrafters/rational-emacs/issues/33

daviwil:
Thanks a bunch!  If people ask enough for this kind of functionality we could certainly adopt your commands, but probably better to leave them out for now.

PR #42, opened 2022-02-17
Add headers to modules

jeffbowman:
This will make them match usage of the skeleton when additional
modules are created, so they all look consistent. Some minor attempt
at words in the "Commentary" section was made, however, more could be
done to provide better documentation for each module.


daviwil:
Awesome, thanks Jeff!

Issue #44, opened 2022-02-18
make doom-modeline optional


erikLundstedt:
im playing around with [powerline](https://github.com/milkypostman/powerline/tree/master) and similar packages and as far as i understand it is much closer to vanilla emacs modeline while allowing for lots of customization

im also thinking about playing around with [telephone-line](https://github.com/dbordak/telephone-line/tree/master) which seems to do ruffly the same

daviwil:
@ajxn Honestly, I have a similar concern.  I think it would be nice if we could make our own nice looking, minimal mode line configuration that doesn't depend on an external package.  Mode line customization is a little harder to understand at first compared to general Emacs config, so it might be nice to have our own mode line here to serve as an example, as you suggested.

I made a recommendation in #45 to introduce a `rational-doom-modeline.el` instead, but I think what we really need first is our own `rational-modeline.el` that doesn't require an external package.

erikLundstedt:
> @ajxn Honestly, I have a similar concern. I think it would be nice if we could make our own nice looking, minimal mode line configuration that doesn't depend on an external package. Mode line customization is a little harder to understand at first compared to general Emacs config, so it might be nice to have our own mode line here to serve as an example, as you suggested.
> 
> I made a recommendation in #45 to introduce a `rational-doom-modeline.el` instead, but I think what we really need first is our own `rational-modeline.el` that doesn't require an external package.

I know im late (by several months) to this, but the implementation I have(its now in my own config repo) allows the user to pick between the existing options(powerline and doom are the ones currently implemented) im thinking of this as more similar to the eglot vs LSP discussion,(powerline being lighter and more emacs-native) at least until we can make something fully custom, if at all.
my module currently allows the user to chose what modeline to use without having to write the config themselves if they dont want to, so merging it shouldn't be too hard

if someone here could help me get `crafted-modeline.el` up to standards that would be great as im not super comfortable in elisp and ive mainly done 1-person git projects where my code is mostly used by myself and where having unnecessary files isn't an issue. i`m also not too comfortable with github as im a gitlab user most of the time

jeffbowman:
As requested on #246, Here is a repost of the comment from closing this issue.

[...]

PR #45, opened 2022-02-18
added variable to make doom-modeline optional

erikLundstedt:
this closes issue #44

Signed-off-by: Erik Lundstedt <erik@lundstedt.it>

### DIFF
--- a/modules/rational-ui.el
+++ b/modules/rational-ui.el
@@ -13,7 +13,6 @@
 ;;; Code:
 
 (straight-use-package 'all-the-icons)
-(straight-use-package 'doom-modeline)
 (straight-use-package 'doom-themes)
 (straight-use-package 'elisp-demos)
 (straight-use-package 'helpful)
@@ -29,14 +28,23 @@ Use a plist with the same key names as accepted by `set-face-attribute'.")
 
 ;;;; Mode-Line
 
-;; Start up the modeline after initialization is finished
-(add-hook 'after-init-hook 'doom-modeline-init)
 
-;; Configure `doom-modeline'
-(customize-set-variable 'doom-modeline-height 15)
-(customize-set-variable 'doom-modeline-bar-width 6)
-(customize-set-variable 'doom-modeline-minor-modes t)
-(customize-set-variable 'doom-modeline-buffer-file-name-style 'truncate-except-project)
+(defcustom rational-ui-use-doom-modeline t
+  "weather or not to use doom-modeline, on by default"
+:type 'boolean
+  )
+
+(when rational-ui-use-doom-modeline
+  (straight-use-package 'doom-modeline)
+  ;; Start up the modeline after initialization is finished
+  (add-hook 'after-init-hook 'doom-modeline-init)
+
+  ;; Configure `doom-modeline'
+  (customize-set-variable 'doom-modeline-height 15)
+  (customize-set-variable 'doom-modeline-bar-width 6)
+  (customize-set-variable 'doom-modeline-minor-modes t)
+  (customize-set-variable 'doom-modeline-buffer-file-name-style 'truncate-except-project)
+)
 
 ;;;; Help Buffers
 


modules/rational-ui.el
```diff
@@ -29,14 +28,23 @@ Use a plist with the same key names as accepted by `set-face-attribute'.")
...
 
-;; Configure `doom-modeline'
-(customize-set-variable 'doom-modeline-height 15)
-(customize-set-variable 'doom-modeline-bar-width 6)
-(customize-set-variable 'doom-modeline-minor-modes t)
-(customize-set-variable 'doom-modeline-buffer-file-name-style 'truncate-except-project)
+(defcustom rational-ui-use-doom-modeline t
+  "weather or not to use doom-modeline, on by default"
+:type 'boolean
+  )
```
angrybacon, line 35:
```suggestion
(defcustom rational-ui-use-doom-modeline t
  "Wether to use doom-modeline."
  :type 'boolean)
```

modules/rational-ui.el
```diff
@@ -29,14 +28,23 @@ Use a plist with the same key names as accepted by `set-face-attribute'.")
...
+  (straight-use-package 'doom-modeline)
+  ;; Start up the modeline after initialization is finished
+  (add-hook 'after-init-hook 'doom-modeline-init)
+
+  ;; Configure `doom-modeline'
+  (customize-set-variable 'doom-modeline-height 15)
+  (customize-set-variable 'doom-modeline-bar-width 6)
+  (customize-set-variable 'doom-modeline-minor-modes t)
+  (customize-set-variable 'doom-modeline-buffer-file-name-style 'truncate-except-project)
+)
```
angrybacon, line 47:
```suggestion
(when rational-ui-use-doom-modeline
  (straight-use-package 'doom-modeline)
  (add-hook 'after-init-hook 'doom-modeline-init)
  (customize-set-variable 'doom-modeline-bar-width 6)
  (customize-set-variable 'doom-modeline-buffer-file-name-style 'truncate-except-project)
  (customize-set-variable 'doom-modeline-height 15)
  (customize-set-variable 'doom-modeline-minor-modes t))
```

angrybacon (changes requested):

erikLundstedt:
> Choice of mode line customization is a pretty big one, on the order of choosing your preferred font and color theme.  
> 
> I'm starting to wonder if we should configure the mode line somewhere else, like in a separate module that can be included only if you want mode line customizations.  That would make it easier to have multiple `rational-doom-modeline.el` etc configurations which don't require configuration variables to decide whether they're turned on or off.
> 
> What do you think?

I don't know

I was planning on making a pr with the telephone-line modeline config I've got where you run a function to use different configurations/defaults and the user get to start the modeline themselves as you need to do the config before starting the modeline with `telephone-line-mode` or something similar

erikLundstedt:
> Choice of mode line customization is a pretty big one, on the order of choosing your preferred font and color theme.  
> 
> I'm starting to wonder if we should configure the mode line somewhere else, like in a separate module that can be included only if you want mode line customizations.  That would make it easier to have multiple `rational-doom-modeline.el` etc configurations which don't require configuration variables to decide whether they're turned on or off.
> 
> What do you think?

Actually
That sounds pretty good
Because then I can write (parts of) the telephone-line module

modules/rational-ui.el
```diff
@@ -29,14 +28,18 @@ Use a plist with the same key names as accepted by `set-face-attribute'.")
...
 
-;; Start up the modeline after initialization is finished
-(add-hook 'after-init-hook 'doom-modeline-init)
 
-;; Configure `doom-modeline'
-(customize-set-variable 'doom-modeline-height 15)
-(customize-set-variable 'doom-modeline-bar-width 6)
-(customize-set-variable 'doom-modeline-minor-modes t)
-(customize-set-variable 'doom-modeline-buffer-file-name-style 'truncate-except-project)
+(defcustom rational-ui-use-doom-modeline t
```
bkaestner, line 32: I don't concur with setting a variable that controls an optional feature to `t` by default.

bkaestner (changes requested):
The current commit history is a bit of a mess. While the intend is fine, can you please clean it up? Given the contents of your changes, you could squash them into a single commit.

modules/rational-ui.el
erikLundstedt: > I don't concur with setting a variable that controls an optional feature to t by default.

thats to not break existing configs
i could change that i guess

erikLundstedt:
> The current commit history is a bit of a mess. While the intend is fine, can you please clean it up? Given the contents of your changes, you could squash them into a single commit.

I am not sure on how to do that as ive only ever used git(gitlab) for my own code where messy history isnt a problem

angrybacon:
> I am not sure on how to do that as ive only ever used git(gitlab) for my own code where messy history isnt a problem

It is always an issue, it's just that GitHub (and GitLab) will optionally let you squash commits while merging once the PR is accepted. That being said, squashing yourself is the conventional thing to do and it helps codeowners review changes more easily without carrying the minor details we talked about here

erikLundstedt:
> > I am not sure on how to do that as ive only ever used git(gitlab) for my own code where messy history isnt a problem
> 
> It is always an issue, it's just that GitHub (and GitLab) will optionally let you squash commits while merging once the PR is accepted. That being said, squashing yourself is the conventional thing to do and it helps codeowners review changes more easily without carrying the minor details we talked about here
> 
> To squash your commits into one, you can pull your branch locally as per latest, and if you're using Magit, ri from the status buffer is probably the easiest

the thing now is that there have been changes made afterwards


im gonna try

PR #48, opened 2022-02-18
Initialize evil collection

ThangaAyyanar:
Quitting using q in overlay buffer doesn't work in the stream so i found the issue and fixed it


daviwil (approved):
That makes sense, thanks a lot for fixing it!

PR #51, opened 2022-02-19
Add support for customizing modes to display line numbers. 

jeastman:
Adds two additional custom variables and support for updating
the display of line numbers based on a change to the
`rational-ui-display-line-numbers` variable.

Added the 'rational' group to each of the custom variables.

Supporting `:set` on the `rational-ui-display-line-numbers`  allows someone to specify:

    (custom-set-variables '(rational-ui-display-line-numbers t))

after `(require 'rational-ui)` and still have the intended effect.

### DIFF
--- a/modules/rational-ui.el
+++ b/modules/rational-ui.el
@@ -64,7 +64,8 @@ Use a plist with the same key names as accepted by `set-face-attribute'.")
 (defcustom rational-ui-line-numbers-disabled-modes
   '(org-mode)
   "Modes which should not display line numbers.
-Modes deriverd from the modes defined in `rational-ui-line-number-enabled-modes', but should not display line numbers."
+Modes deriverd from the modes defined in
+`rational-ui-line-number-enabled-modes', but should not display line numbers."
   :type 'list
   :group 'rational)
 
--- a/modules/rational-ui.el
+++ b/modules/rational-ui.el
@@ -86,11 +86,9 @@ Used as hook for modes which should not display line numebrs."
   (if rational-ui-display-line-numbers
       (progn
         (dolist (mode rational-ui-line-numbers-enabled-modes)
-          (message "enabling line numbers for %s" mode)
           (add-hook (intern (format "%s-hook" mode))
                     #'rational-ui--enable-line-numbers-mode))
         (dolist (mode rational-ui-line-numbers-disabled-modes)
-          (message "disabling line numbers for %s" mode)
           (add-hook (intern (format "%s-hook" mode))
                     #'rational-ui--disable-line-numbers-mode))
         (setq-default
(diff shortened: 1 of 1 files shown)


ajxn:
I would suggest to add line numbers to `prog-mode-hook` and not to `text-mode-hook`, maybe as default.  Still have list of modes, of course.

Would make more sense for me, at least.

jeastman:
The original code looked like:

    (add-hook 'conf-mode-hook #'display-line-numbers-mode)
    (add-hook 'prog-mode-hook #'display-line-numbers-mode)
    (add-hook 'text-mode-hook #'display-line-numbers-mode)

Which is why I included text mode as well. I agree it is probably better to leave that one off of the default list.

modules/rational-ui.el
```diff
@@ -55,18 +55,63 @@ Use a plist with the same key names as accepted by `set-face-attribute'.")
...
+(defcustom rational-ui-line-numbers-enabled-modes
+  '(conf-mode prog-mode)
+  "Modes which should display line numbers."
+  :type 'list
+  :group 'rational)
+
+(defcustom rational-ui-line-numbers-disabled-modes
+  '(org-mode)
+  "Modes which should not display line numbers.
+Modes deriverd from the modes defined in
```
daviwil, line 67:
```suggestion
Modes derived from the modes defined in
```

modules/rational-ui.el
```diff
@@ -55,18 +55,63 @@ Use a plist with the same key names as accepted by `set-face-attribute'.")
...
+
+Used as hook for modes which should not display line numebrs."
+  (display-line-numbers-mode -1))
+
+(defun rational-ui--update-line-numbers-display ()
+  "Update configuration for line numbers display."
+  (if rational-ui-display-line-numbers
+      (progn
+        (dolist (mode rational-ui-line-numbers-enabled-modes)
+          (message "enabling line numbers for %s" mode)
```
daviwil, line 89: Was the message for debugging purposes?  Might be good to remove it for normal use

modules/rational-ui.el
```diff
@@ -55,18 +55,63 @@ Use a plist with the same key names as accepted by `set-face-attribute'.")
...
+          (add-hook (intern (format "%s-hook" mode))
+                    #'rational-ui--enable-line-numbers-mode))
+        (dolist (mode rational-ui-line-numbers-disabled-modes)
+          (message "disabling line numbers for %s" mode)
```
daviwil, line 93: Same here

daviwil (approved):
Great improvement!  Just a couple tweaks and we can merge it

modules/rational-ui.el
jeastman: 2nd one of these... I obviously cannot spell.

daviwil:
Thanks a bunch!

Issue #37, opened 2022-02-16
theme and font settings don't work starting emacs as a daemon


fnt400:
I tried to set my theme (prot's one) and font (iosevka) as raccomended in the README file. 
It works fine starting emacs normally, but if I start emacs as daemon (or with systemd) and than I create a new frame with emacsclient -c  it has the blueish default theme and another font.

jeffbowman:
Try setting those in the `rational-emacs/early-config.el` file. See if that helps.

In general, setting fonts and frame based configuration in early-init (or early-config.el in the case of rational-emacs), is a better idea because they get set before the frame exists so the springy resizing thing does not happen when the frame becomes visible. In daemon mode, that *should* help with getting the emacs client frame looking the way you want.

fnt400:
> @fnt400 Try setting those in the `rational-emacs/early-config.el` file. See if that helps.
thankyou for the advice. I tried but it didn't work. The font do not change when starting as a daemon. 
The theme setting, instead, gives an error because it can't find the theme itself (probably it can be fixed adding a load-path but I didn't try)

jeffbowman:
@fnt400 The reason you are seeing the blue color has to do with the default configuration in the rational-emacs early-init.el file, specifically these 2 lines:

https://github.com/SystemCrafters/rational-emacs/blob/master/early-init.el#L29-L30

which set the default foreground and background colors.

In your `rational-config-path/early-config.el` you need to add this configuration to get rid of it:

```
(setf default-frame-alist (assq-delete-all 'background-color default-frame-alist))
(setf default-frame-alist (assq-delete-all 'foreground-color default-frame-alist))
```

Then your code to load a them like this `(load-theme 'modus-vivendi t)` (probably in your `rational-config-path/config.el` file)  should work.

fnt400:
Great! Now the theme starts correctly! Unfortunately the font setting doesn't. I still have to set it manually when I create the first frame (but that's not a big problem).

jeffbowman:
This is what my `early-config.el` file has. the font gets set correctly as does the theme. Does this help you?

https://gist.github.com/jeffbowman/7d0e20df9321aaa28468c2a32503d322

fnt400:
Yes!!! That did the trick! I was using the function set-face-attribute, instead of custom-set-faces.
Thank you very much!!

PR #53, opened 2022-02-19
Populate content into example-config.el

jeastman:
The example-config.el file had no content. This PR simply populates it with some comments and the example from README.org.

### DIFF
--- a/example-config.el
+++ b/example-config.el
@@ -2,7 +2,7 @@
 
 ;;; Commentary:
 ;;
-;; Rational Emacs supports user customization throug a `config.el' file
+;; Rational Emacs supports user customization through a `config.el' file
 ;; similar to this one.  You can copy this file as `config.el' to your
 ;; Rational Emacs configuration directory as an example.
 ;;
--- a/example-config.el
+++ b/example-config.el
@@ -1,0 +1,31 @@
+;;; example-config.el -- Example Rational Emacs user customization file -*- lexical-binding: t; -*-
+
+;;; Commentary:
+;;
+;; Rational Emacs supports user customization throug a `config.el' file
+;; similar to this one.  You can copy this file as `config.el' to your
+;; Rational Emacs configuration directory as an example.
+;;
+;; In your configuration you can set any Emacs configuration variable, face
+;; attributes, themes, etc as you normally would.
+;;
+;; See the README.org file in this repository for additional information.
+
+;;; Code:
+(require 'rational-defaults)
+(require 'rational-screencast)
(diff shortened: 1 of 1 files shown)


daviwil (approved):
Looks great, thanks a lot!  Just one little suggested change

example-config.el
```diff
@@ -0,0 +1,31 @@
...
+(require 'rational-defaults)
+(require 'rational-screencast)
+(require 'rational-ui)
+(require 'rational-editing)
+(require 'rational-evil)
+(require 'rational-completion)
+(require 'rational-windows)
+
+;; Set further font and theme customizations
+(set-face-attribute 'default nil
```
jeffbowman, line 24: A better way would be to define the default face in early-config.el. The reason is, it gets set and the frame size can be calculated based on the font size before it is rendered, which avoids an annoying resizing behavior. Also why we turn off the toolbar and menubar there (emacs avoids having to compute a size for a frame containing them only to turn them off once the frame is rendered, which then causes the frame to suddenly shrink).

example-config.el
```diff
@@ -0,0 +1,31 @@
...
+                  :font "JetBrains Mono"
+                  :weight 'light
+                  :height 185)
+
+(load-theme 'doom-snazzy t)
```
jeffbowman, line 29: This may or may not work as expected since we explicitly set the foreground and background colors on the `default-frame-alist`. See #37 for discussion and resolution.

jeffbowman (commented):
Good stuff, didn't realize I started a review, thought I was just adding a comment (d'oh!)... So, don't mind the review thing, the comments my be worth thinking about (or not as the case may be :) )

example-config.el
daviwil: Ouch, this shouldn't be necessary to theme things correctly.  Maybe we should remove the default background color setting?  The thing I'm trying to avoid with that is a flash of white when the user starts up the configuration, but maybe the current fix is more problematic than necessary to accomplish it...

example-config.el
daviwil: Yeah, you make a good point.  It's probably better to show an example of setting the font in `early-config.el`.  That would have solved the problem we were having on stream yesterday where my font was tiny until the full config loaded up :)

example-config.el
ajxn: If we need `custom-file' loaded early, then we probably need to move that into `early-init.el' from beginning of `init.el'  too.

Or would that be a problem? It shouldn't be, should it?

PR #47, opened 2022-02-18
make use of use-short-answers

jeastman:
Fix for issue #43 

Utilize the `use-short-answers` variable, if available.

Issue #43, opened 2022-02-17
use-short-answer vs. y-or-n-p


jeastman:
Emacs 28.1 has the variable use-short-answers which can be used. This could be used instead of resetting `yes-or-no-p` know this is a newer feature, but it may be a good idea to begin using the facilities provided by Emacs.

To be backward compatible, you will still need to test for the presence of the variable.

    (if (boundp 'use-short-answers)
        (setq use-short-answers t)
      (fset 'yes-or-no-p 'y-or-n-p))

jeastman:
I should note that the docs say:

    We recommend against setting this variable non-nil, because yes-or-no-p
    is intended to be used when users are expected not to respond too
    quickly, but to take their time and perhaps think about the answer.

But I think that warning applies to replacing `yes-or-no-p` with `y-or-n-p` as well.

jeastman:
See additional PR #54

jeffbowman:
I'm in such a habit of typing "yes" or "no" that turning that to 'y' or 'n' for me breaks things... weird things happen... I type 'y' the prompt does something, then I get "es" to show up in a buffer... or suddenly my dired sorts differently... So, I personally prefer the longer form. Interesting insight @jeastman , I had not considered it from that perspective.

daviwil:
I have the same problem, I'm very used to typing `yes` in many places.  Maybe we could have a section of "suggested settings" in the README to point the user to some things that they might be interested to try?  If we make it harder to just copy/paste a whole block of settings then it's possible that they would actually think about whether they want this behavior or not.

As a side note, looking at the docs for `use-short-answers` pointed me to the `read-answers` function which is much better for what I was trying to do on stream for prompting the user about pulling updates.

jeffbowman:
> Maybe we could have a section of "suggested settings" in the README to point the user to some things that they might be interested to try?

[...]

PR #54, opened 2022-02-19
Use alias instead of fset

jeffbowman:
fset is really difficult to revert in the event the user wishes to use longer answers (ie "yes"/"no") for confirm/negate prompts. This patch uses advice and includes logic from #47 making use of `use-short-answers` when it is available.


daviwil:
I'll wait until we conclude the discussion in #43 before I merge this.

PR #57, opened 2022-02-19
Added detection of wether it is running guix's emacs

abarocio80:
A new function (`rational-runs-guix-emacs-p') to detect if we are running an emacs in the /gnu/store/.
Also sets the default value of `rational-prefer-guix-packages' to the result of the new added function.

Of course, this could be overriden on `early-config.el', if needed.

### DIFF
--- a/early-init.el
+++ b/early-init.el
@@ -68,7 +68,7 @@
               (equal system-type 'windows-nt))
     ;; Since there is no windows implementation of guix
     (string-prefix-p "/gnu/store/"
-                     (file-chase-links
+                     (file-truename
                       (executable-find
                        (car command-line-args))))))
 


early-init.el
```diff
@@ -62,10 +62,21 @@
...
+                       (car command-line-args))))))
+
+(defvar rational-prefer-guix-packages (rational-runs-guix-emacs-p)
   "If t, expect packages to be installed via Guix by default.")
 
 ;; Load the early config file if it exists
 (let ((early-config-path (expand-file-name "early-config.el" rational-config-path)))
   (when (file-exists-p early-config-path)
     (load early-config-path nil 'nomessage)))
+(executable-find (car command-line-args))
```
daviwil, line 82: Probably not meant to be here?

abarocio80 (commented):
Done. Both functions work with emacs 27.2 (current guix version) out of the box.

early-init.el
```diff
@@ -62,10 +62,21 @@
...
 
-(defvar rational-prefer-guix-packages nil
+(defun rational-runs-guix-emacs-p ()
+  "Verifies if the running emacs executable is under the `/gnu/store/' path."
+  (unless (or (equal system-type 'ms-dos)
+              (equal system-type 'windows-nt))
+    ;; Since there is no windows implementation of guix
+    (string-prefix-p "/gnu/store/"
+                     (file-chase-links
+                      (executable-find
```
daviwil, line 72: This is much better, thanks a lot!

early-init.el
```diff
@@ -62,10 +62,21 @@
 (unless (file-exists-p rational-config-path)
   (mkdir rational-config-path t))
 
-(defvar rational-prefer-guix-packages nil
+(defun rational-runs-guix-emacs-p ()
```
daviwil, line 65: Maybe `rational-using-guix-emacs-p` is slightly more clear?

daviwil (commented):
Almost perfect, just a couple more things

early-init.el
abarocio80: Oh, right!

early-init.el
abarocio80: OK. Should I also change `file-chase-links` to `file-truename`? In the documentation I understand that the latter does a better job with symlinks in the directories of the path.

daviwil: Yep, that is probably the better choice as long as it doesn't prompt the user to follow the symlink.  My guess is that it doesn't, you'll find out right away as soon as you test it :)

abarocio80: I will...

daviwil:
Awesome, thanks a lot Alejandro!

PR #58, opened 2022-02-19
Emacs init time and load with`noerror' 

ajxn:
Store and load custom variables in "emacs-custom.el" instead of in "init.el".

Make load ignore errors, like non existing files with `noerror'.

### DIFF
--- a/init.el
+++ b/init.el
@@ -32,7 +32,7 @@
         'silent 'inhibit-cookies)
       (goto-char (point-max))
       (eval-print-last-sexp)))
-  (load bootstrap-file nil 'nomessage))
+  (load bootstrap-file 'noerror 'nomessage))
 
 (defun rational-ensure-package (package &optional args)
   "Ensure that PACKAGE is installed on the system, either via
--- a/init.el
+++ b/init.el
@@ -1,5 +1,10 @@
 ;;; init.el -*- lexical-binding: t; -*-
 
+;; Don't save customize variables in `init.el', save them in
+;; "emacs-custom.el" instead. Don't bail out if file doesn't exist.
+(setq custom-file (locate-user-emacs-file "emacs-custom.el"))
+(load custom-file 'noerror)
+
 ;; Profile emacs startup
 (add-hook 'emacs-startup-hook
           (lambda ()
--- a/init.el
+++ b/init.el
@@ -58,8 +58,7 @@ straight.el or Guix depending on the value of
 (mkdir rational-config-var-directory t)
 
 ;; Load the user configuration file if it exists
-(when (file-exists-p rational-config-file)
-  (load rational-config-file nil 'nomessage))
+(load rational-config-file 'noerror 'nomessage)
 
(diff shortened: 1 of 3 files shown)


init.el
```diff
@@ -1,12 +1,15 @@
...
+(load custom-file 'noerror 'nomessage)
+
 ;; Profile emacs startup
 (add-hook 'emacs-startup-hook
           (lambda ()
             (message "Rational Emacs loaded in %s."
-                     (format "%.2f seconds"
-                             (float-time
-                              (time-subtract after-init-time before-init-time))))))
+                     (emacs-init-time))))
```
bkaestner, line 12: Please squash the commits on this section.

init.el
```diff
@@ -29,7 +32,7 @@
         'silent 'inhibit-cookies)
       (goto-char (point-max))
       (eval-print-last-sexp)))
-  (load bootstrap-file nil 'nomessage))
+  (load bootstrap-file 'noerror 'nomessage))
```
bkaestner, line 35: This ***must*** return an error, otherwise `straight.el`'s bootstrap file is missing and most of the rest of this configuration cannot work.

While the previous lines are explicitly setting up `straight.el`, it might still fail if the `user-emacs-directory` is read-only or similar. Or `install.el` might contain `nil` due to a broken repository. In either cases, we don't have `straigth.el` available and *really* should know about that.

Please revert this change.

init.el
```diff
@@ -1,12 +1,15 @@
 ;;; init.el -*- lexical-binding: t; -*-
 
+;; Don't save customize variables in `init.el', save them in
+;; "emacs-custom.el" instead. Don't bail out if file doesn't exist.
+(setq custom-file (locate-user-emacs-file "emacs-custom.el"))
+(load custom-file 'noerror 'nomessage)
+
```
bkaestner, line 7: This change seems only slightly related to the other changes and is also opinionated. I'd move that one in another pull request, to be honest.

bkaestner:
There are a lot of squashes and merges in this PR, as well as reverted changes and duplicate commit messages. Can you please clean up the PR/branch? Given the changes in the files, I'd expect maybe three, easily revertible commits, e.g.

- use `'noerror` on optional files
- use `emacs-init-time` instead of manual calculation
- set and load `custom-file`

The current set of 12 commits are a little bit of a mess in my opinion. Feel free to ping me if I shall clean that up for you :)

bkaestner (changes requested):
There are a lot of squashes and merges in this PR, as well as reverted changes and duplicate commit messages. Can you please clean up the PR/branch? Given the changes in the files, I'd expect maybe three, easily revertible commits, e.g.

- use `'noerror` on optional files
- use `emacs-init-time` instead of manual calculation
- set and load `custom-file`

The current set of 12 commits are a little bit of a mess in my opinion. Feel free to ping me if I shall clean that up for you :)

That being said, great idea about using `load`'s `'noerror` parameter! I thought that was already used, but I might misremember something here. And also good catch on `(emacs-init-time)`!

PR #65, opened 2022-02-20
Highlight current line in prog mode

jiploading:


modules/rational-ui.el
```diff
@@ -127,5 +127,11 @@ Used as hook for modes which should not display line numebrs."
 				                     recenter-top-bottom other-window))
   (advice-add command :after #'pulse-line))
 
+;; Italic comments and keywords in server-mode
+(set-face-attribute 'font-lock-comment-face nil
+  :slant 'italic)
+(set-face-attribute 'font-lock-keyword-face nil
+  :slant 'italic)
+
```
jeastman, line 135: I would suggest removing the Italic comments section from this PR.

Italicizing comments is something often handled by a theme. It does not seem right that we would override the user's theme preference by forcing italic comments here.

jiploading: How about we give user a choice. Whether they want to use italic comments or not?

japhir:
Oh shouldn't this hl-line-mode then replace the native beacon mode that I added at some point?

modules/rational-ui.el
japhir: I wouldn't add this here, would be something to put in config.el or inherit from the theme

jeastman:
> Oh shouldn't this hl-line-mode then replace the native beacon mode that I added at some point?

Maybe "highlighting" becomes its own feature?

daviwil:
Do we want the line highlighting feature added?  Use the reaction emojis for 👍🏻 and 👎🏻 to vote on this comment!

onetom:
i only turn the `hl-line-mode` off in `markdown-view-mode` / `gfm-view-mode`.
combined with the modus themes, the highlighting is very subtle, but even the out of the box green color is alright.

PR #66, opened 2022-02-20
Starting configuration for project management

abarocio80:
I have noticed that there is nothing to point the `project-list-file` variable to a file in the `rational-config-path`, because the default is to lookup to `user-emacs-directory`, which I belive is not quite right, since that file is a user's customization.

Maybe some other configurations could be added here.


daviwil:
Thanks a lot!

PR #67, opened 2022-02-20
Fix the default skeleton for modules

abarocio80:
The default skeleton is generating

``` lisp
;; Commentary
```

instead of

``` lisp
;;; Commentary:
```

### DIFF
--- a/init.el
+++ b/init.el
@@ -74,7 +74,7 @@ straight.el or Guix depending on the value of
 
 ;; Author: System Crafters Community
 
-;; Commentary
+;;; Commentary:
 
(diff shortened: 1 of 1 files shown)


jeastman:
I noticed this also and fixed this as part of PR #68

abarocio80:
I don't see the change in the PR.

jeastman:
Noted. I had only fixed the ':'. I will update with the third ';' as well.

jeastman:
As #68 is a more dramatic change, this could be merged first and #68 could be rebased on top of it.

Issue #59, opened 2022-02-19
Follow Emacs guidelines/best practices in documentation/modules


bkaestner:
As per [`(info "(elisp) Tips for Documentation Strings")`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html), the function documentations should use the imperative verb forms:

> For consistency, phrase the verb in the first sentence of a function’s documentation string as an imperative—for instance, use “Return the cons of A and B.” in preference to “Returns the cons of A and B.” 

For example, within `rational-updates.el`, the documentation for `rational-updates-check-for-latest` might be changed into:

```diff
 (defun rational-updates-check-for-latest ()
-   "Fetches the latest Rational Emacs commits from GitHub and
- notifies you if there are any updates."
+   "Fetch the latest Rational Emacs commits. Show changes if there are any."
```

There are also many more tips in the given info page. We might consider `checkdoc` and/or `package-lint` in the CI.

daviwil:
Either of those options in CI would be great!

bkaestner:
Do you consider each module its own package? Or should the whole `rational-emacs` be considered a package? Depending on that we might need an additional `rational-pkg.el` in `modules/` or need to fix a lot of lints, see https://github.com/bkaestner/rational-emacs/actions.

daviwil:
Good question, I would technically consider each to be their own package.  One of the goals was that each module should, in theory, be usable outside of Rational Emacs, though I don't know how realistic that will be in practice.  What do you think is the best approach?

bkaestner:
I'm not sure yet. At the moment, `rational` is not self-contained, as it needs/has some functionality in `init.el` and `early-init.el`. Without the appropriate user configuration, none of the packages can be loaded/byte compiled.

We probably need a `rational.el` base module first, that provides the functionality from the current `init.el`/`early-init.el`. On the one hand, that would also enable installation as a package from MELPA or similar. On the other hand, that's somewhat contradictory to creating a base distribution.

Hm. Maybe we could slim down the current `{early-}init.el`, put those functions into their own modules and load those modules instead? The biggest challenge might be the byte-compilation against packages that are unknown at byte-compile time if we don't take care.

daviwil:
Maybe a new `rational-core.el` module that has the core functionality that is needed at init time?  It would definitely make sense to pull all of the functions and variables we define out of the `init.el` files.

daviwil:
[...]

PR #68, opened 2022-02-20
Create rational-emacs module for initialization.

### LINKED ISSUE #62: Define customization group
There needs to be a `defgroup` for the `rational` customize group somewhere to link the custom variables into.

I was thinking something along the lines of the following:

    (defgroup 'rational
        "A sensible starting point for hacking your own Emacs configuration."
        :group 'emacs)

The problem is that we need some place to define this which is guaranteed to be loaded. A discussion of this common source is also discussed in #59.

jeastman:
Attempts to address issue #59.
Fixes #62.

The contents of early-init.el and init.el have moved to functions
in the rational-emacs.el file.

early-init.el and init.el are updated to call the appropriate
functions in this new module.

All functions are updated to reflect the module name 'rational-emacs'.


jeastman:
Note this PR will conflict with other unmerged PRs due to moving the content of init.el.

- The fix suggested in #67 is included in this PR (I noticed the same issue). @abarocio80
- PR #58 appeared to be under discussion, so I did not include the proposed changes. @ajxn
- PR #14 appeared to be under discussion, so I did not include the proposed changes. @mfxarc

jeastman:
Updated skeleton (and file header) based on comments from #67

ajxn:
I closed #58 as I didn't had time to fix it (and it was a mess when it came to commits), so that isn't a problem any more.

jeastman:
Closing this issue as there was too much covered in it and the discussion in #59 is not completed.

PR #74, opened 2022-02-20
A module for autocompile emacs lisp code

abarocio80:
# Description
This module is intended for automate the compilation of emacs-lisp code on some files.

- Files in the `modules` directory
- Files in the `rational-config-path` directory
- Init files
- Specific directories
- And all `emacs-lisp-mode` buffers after save.

All is configurable via simple variables.

# Possible enhancement
- Define clearly when to autocompile the modules and/or the init files.
  For the moment, I am using `emacs-startup-hook`, but this could be changed.
- Change variables to customizations.
- Implement better interactive functions.
- Autocompile on save only on specific directories (and subdirectories?).

### DIFF
--- a/modules/rational-autocompile-el.el
+++ b/modules/rational-autocompile-el.el
@@ -69,7 +69,7 @@ directory will be compiled, but not it's subdirectories."
       (cl-return nil))
   (let ((file (buffer-file-name b)))
     (when file
-      (rational-autocompile-el-file file))))
+      (rational-autocompile-el-file file)))))
 
 ;; A function to compile a specific directory
 (defun rational-autocompile-el-directory (d)
--- a/modules/rational-autocompile-el.el
+++ b/modules/rational-autocompile-el.el
@@ -1,0 +1,154 @@
+;;;; rational-autocompile-el.el --- Autocompile emacs-lisp code  -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2022
+;; SPDX-License-Identifier: MIT
+
+;; Author: System Crafters Community
+
+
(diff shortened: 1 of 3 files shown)


modules/rational-autocompile-el.el
```diff
@@ -0,0 +1,154 @@
...
+        (if (file-directory-p source)
+            (byte-recompile-directory source 0)
+          (byte-recompile-file source nil 0))))))
+
+;; A function to compile the buffer's file
+(defun rational-autocompile-el-buffer (&optional b)
+  "Compiles (native or byte-code) the file of buffer B."
+  (when b
+    (unless (bufferp b)
+      return nil))
```
jeastman, line 69: This should be `cl-return` instead of return.

abarocio80: Oh, right! It also needs parens.

abarocio80: Done!

abarocio80:
Missing parentesis fixed.

jeastman (approved):

abarocio80:
I have changed the name of the package, as @daviwil suggested [in his review video](https://youtu.be/f8LNWLWOIXI?t=2985).

abarocio80:
I'm still not happy with the module name.  Any suggestions?

abarocio80:
I don't want to convey that it will autocompile any code.

jeffbowman:
Given how it is written, I'm okay with the name. Maybe `rational-compile` would be better for you?

abarocio80:
Yea, maybe.

Issue #31, opened 2022-02-15
No command history in evil-mode from previous session


rotlaus:
When typing : to start a command after starting emacs i get no command history to scroll through from the previous session.
Message Buffer says 
`previous-matching-history-element: No earlier matching history item`

I do get a history of commands from the current session, but after a fresh emacs start it's empty.

rotlaus:
Enabling savehist-mode fixes my problem. 
I could make a pull request. Should it be enabled always ( in rational-defaults) or only for evil mode (in rational-evil)?

jeffbowman:
I don't preserve kill-ring, search, et.al. like that. I actually didn't know that was a thing! So, I learned something. Neat!

PR #75, opened 2022-02-21
Add module rational-persistence.el

hey-jeldo:
Some settings related to persistence of emacs over different sessions.
I like these a lot and didn't know about them for the longest time.
Maybe this is a way to expose these settings for users that are interested in this behaviour.


modules/rational-persistence.el
```diff
@@ -0,0 +1,35 @@
...
+(setq savehist-save-minibuffer-history t)
+  (setq savehist-additional-variables
+        '(kill-ring
+          search-ring
+          regexp-search-ring
+          last-kbd-macro
+          kmacro-ring
+          shell-command-history))
+
+(savehist-mode)
```
jeastman, line 27: `savehist-mode` is also activated in `rational-defaults.el` (see L48-49). You will want to remove it from that module as part of this PR.

There is some additional discussion in Issue #31 which may be relevant as well.

japhir: Are some of these options also available via setq-default or customize-set-variable?

hey-jeldo: > Are some of these options also available via setq-default or customize-set-variable?

I don't know, i remember recently hearing about the difference of these methods (probably in a System Crafters video) but i don't understand it fully just yet. I think since these variables are not buffer local, `setq` and `setq-default` would do the same thing here? `customize-set-variable` i am not sure at all. 😅

jeastman: The biggest difference with `custom-set-variables` is that it will trigger any associated logic with setting the value. If done correctly by the implementor of the `defcustom`, it does not matter when you set the value (e.g. before/after `require`, before/after starting a mode, etc.). If there is no `:set` on the `defcustom`, it is (almost always) equivalent to using `setq-default`.

hey-jeldo: Thank you! That clears things up!

PR #78, opened 2022-02-21
(rational-editing) Add overwrite region and subword-mode

hey-jeldo:
I think having selected text be overwritten is expected behaviour for most people.
Being able to `M-f` and `M-b` through camelCasedWords also seems a reasonable default.

I am quite new to contributing on GitHub (or anywhere for that matter), so if these should rather be Issues that PRs or I miss anything else here, kindly let me know. :)

### DIFF
--- a/modules/rational-editing.el
+++ b/modules/rational-editing.el
@@ -25,5 +25,11 @@
 (electric-pair-mode 1) ; auto-insert matching bracket
 (show-paren-mode 1)    ; turn on paren match highlighting
 
+;; When text is selected, overwrite it by typing
+(delete-selection-mode 1)
+
+;; easilyNavigateCamelCase
+(global-subword-mode 1)
+
 (provide 'rational-editing)
 ;;; rational-editing.el ends here


japhir:
Looks good to me! I have the same in my config.
I also have:
```
  ;; backspace deletes one character instead of one column
  (global-set-key (kbd "DEL") 'backward-delete-char)
```
but I'm not sure that is general enough for inclusion.

ajxn:
I think that the `global-subword-mode` should not be used in text-mode or other computer languages.  As you also has `global-superword-mode`, which suits other languages then Java better.
Better to add `subword-mode` or `superword-mode` to hooks.

daviwil:
I think we should leave out `global-subword-mode` and suggest it perhaps in a documentation file for this project.  I like the `delete-selection-mode` though!

modules/rational-editing.el
```diff
@@ -25,5 +25,11 @@
 (electric-pair-mode 1) ; auto-insert matching bracket
 (show-paren-mode 1)    ; turn on paren match highlighting
 
+;; When text is selected, overwrite it by typing
+(delete-selection-mode 1)
+
+;; easilyNavigateCamelCase
+(global-subword-mode 1)
```
jeastman, line 32: Per @daviwil 's comment, could you remove the `global-subword-mode`?

jeastman (changes requested):

PR #79, opened 2022-02-21
(gitignore) Add rational-emacs/

hey-jeldo:
Since for chemacs users without environment variable, the rational-emacs folder is assumed in the profile, it should be ignored by git.


jeastman (approved):
LGTM

PR #81, opened 2022-02-21
(rational-ui) macOS transparent titlebar

hey-jeldo:
Check for the `system-type` and load `ns-auto-titlebar.el` to make emacs look more native on current macOS versions.

With a light theme:
Screenshot 2022-02-21 at 22 43 41

With a dark theme:
Screenshot 2022-02-21 at 22 44 07

When theme is toggled, so is the titlebar.

### DIFF
--- a/modules/rational-ui.el
+++ b/modules/rational-ui.el
@@ -127,5 +127,12 @@ Used as hook for modes which should not display line numebrs."
 				                     recenter-top-bottom other-window))
   (advice-add command :after #'pulse-line))
 
+
+;;;; Make the titlebar on macOS transparent
+
+(when (eq system-type 'darwin) (progn
+                                 (straight-use-package 'ns-auto-titlebar)
+                                 (ns-auto-titlebar-mode)))
+
 (provide 'rational-ui)
 ;;; rational-ui.el ends here


ajxn:
Just to check, does this work when not in GUI?  That is `emacs -nw`?  If so and with the changes suggested, it looks ok.

I don't belive you need the `progn` in `when`, so you can remove it. 

I also believe the part after the predicate (the body) should start on the next line.

TDiazT:
(This is me just wondering because I'm not familiar with all the details)
> I don't belive you need the progn in when, so you can remove it.

I tried evaluating this without the `progn` part and it actually throws an error (`invalid-function ...`), so maybe it is necessary 🤷 .

ajxn:
> (This is me just wondering because I'm not familiar with all the details)
> 
> > I don't belive you need the progn in when, so you can remove it.
> 
> I tried evaluating this without the `progn` part and it actually throws an error (`invalid-function ...`), so maybe it is necessary shrug .

Try this expression in Emacs `*scratch*` buffer, and end with `C-j` to evaluate it.  `t` means true.

    (when t
        (message "Message one")
        (message "Message two"))

And the result of the last function, "Message two" will be printed.  And if you look in buffer `*Messages*` you'll see these two messages.  If you set first argument to `nil`, that is false, it will just return `nil`.

    (when nil
        (message "one")
        (message "two"))

You get the documentation for `when` if you try C-h f and type when.

But yes, it will also work with a `progn`.

ajxn:
> (This is me just wondering because I'm not familiar with all the details)
> 
> > I don't belive you need the progn in when, so you can remove it.
> 
> I tried evaluating this without the `progn` part and it actually throws an error (`invalid-function ...`), so maybe it is necessary shrug .

Depending on if it is MacOSX in text mode or the graphical version of Emacs, you should compare if `system-type` is `darwin` (for OS type), as you write or `window-system` is `mac` if you run Emacs in graphical mode in MacOSX.
You can use `(eq system-type darwin)` or `(memq window-system '(mac ns x))` if you want to compare a variable to see if it is one of the values in the list.  In this case, I would do `eq` as it should be true for one value.

Yes, this is needed to work good with Mac OSX. So please fix it and the support for Mac OSX will be better.

daviwil:
Thanks!  I agree this would make the UI look nicer for users on macOS, though I wonder if we can avoid pulling in a separate package for this.  Could we just add the real setting to the frame default parameters?

https://github.com/purcell/ns-auto-titlebar/blob/master/ns-auto-titlebar.el#L58

PR #82, opened 2022-02-21
Use package.el to install use-package

jeffbowman:
It seemed odd to use straight to install `use-package`, when the rest
of this module is about using package.el. Update priorities to pull
from gnu/nongnu elpas over melpa per the principles, install
`use-package` from one of the elpas if it is not already
installed. Also do not activate packages when calling
`package-initialize`, prefer to allow `use-package`, `require`, or
`autoload` to activate packages as needed by the user.

### DIFF
--- a/modules/rational-use-package.el
+++ b/modules/rational-use-package.el
@@ -5,14 +5,12 @@
 
 ;; Author: System Crafters Community
 
-;; Commentary
+;;; Commentary:
 
 ;; Built-in `package.el' with `use-package' configuration, including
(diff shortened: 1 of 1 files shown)


japhir:
Looks good to me! :)

hey-jeldo:
Yes, makes sense!

daviwil:
Let's discuss whether we should stick with `straight.el` or switch to `package.el`: https://github.com/SystemCrafters/rational-emacs/issues/94

PR #84, opened 2022-02-22
Configure custom-file location in rational-config-path

### LINKED ISSUE #83: (init.el) move customised variables to their own file
In it's current state those customised variables still go into the init.el file, which is version controlled and can get overwritten by `rational-updates`, if i understand correctly.

Screenshot 2022-02-22 at 20 44 41

jeffbowman:
This will cause emacs to write the `custom-set-variables' code to the
rational-config-path/custom.el file, and should fix #83


jeastman (approved):

ajxn:
The config file also need to be loaded with `load` command.  And that before the first use of custom variables, so possible in the beginning of `init.el` or `early-init.el`.

jeffbowman:
Why does it need to be loaded? Using `customize-set-variable` does the job for setting values throughout the configuration. The `customize-set-variables` section in custom is written there assuming the user used the EZ Config UI to set the values. Since we didn't we don't need to actually `load` the file. Maybe I'm missing something though?

daviwil:
It definitely needs an additional load for the custom config file, Emacs doesn't load it automatically.

jeffbowman:
> It definitely needs an additional load for the custom config file, Emacs doesn't load it automatically.

Then each call to `customize-set-variable` will set the value and then loading the `custom.el` file will set the exact same values again in the `customize-set-variables` form and slows down the start of Emacs. But loading the custom file does no harm.

jeastman:
The `custom.el` file could be pulled in very early to partially solve this. Due to the way custom work, the setting will not be applied until the relative module is required. However, this leaves us at a place where any of the custom values set in the Rational configuration would always override user configuration.

jeffbowman:
> However, this leaves us at a place where any of the custom values set in the Rational configuration would always override user configuration.

Right. Although, this should be a minimal risk and would be problematic if the user decided to edit the `custom-set-variables` form or the `custom-set-faces` form rather than using other configuration forms in either `config.el` or `early-config.el`.

The custom.el only contains 2 blocks of code: these are the `custom-set-variables` form and the `custom-set-faces` form. Any font updates in either `early-init.el` or `early-config.el` would potentially be clobbered by loading `custom.el` later, which sort of means we would need to load this as the last thing in `early-init.el` or provide documentation on how to load it very early in the users `early-config.el` to avoid that situation... and that assumes they don't change things later on in their own configuration for one of the variables. Bottom line is: we just don't need to load it at all, especially since we are (I think) trying to be disciplined about using `customize-set-variable` for `defcustom` defined values. Writing these 2 forms to an alternate file can be considered part of the "no littering" philosophy. This file could even live in the `rational-config-var-directory`.

This file really is for someone who is using the Customization UI to configure Emacs. When using that UI, Emacs needs a way to persist the values so they are loaded the next time Emacs starts. We aren't doing that. We aren't even advocating a user of Rational Emacs _should_ do that. 

All that being said, This code could be in a users `config.el` and (in theory) would be reasonably valid: ( _not tested_ )

    ;; load rational modules, their customize-set-variables will apply "defaults"
    (require 'rational-defaults)
    ;; require other rational modules as they see fit.
    ;; tell emacs where the custom file is
    (customize-set-variable 'custom-file (expand-file-name "custom.el" rational-config-var-directory))
    ;; make sure its on the load path
    (add-to-list 'load-path rational-config-var-directory)
    ;; load it... this further assumes the *only* place a user will update customizable values will be 
    ;; either in the  custom.el file itself or through the Customize UI, so this *must* be the last thing 
    ;; that happens in the configuration.
    (load "custom" t t)
    ;; config.el ends here

jeastman:
> especially since we are (I think) trying to be disciplined about using `customize-set-variable` for `defcustom` defined values.
I agree.

> We aren't doing that. We aren't even advocating a user of Rational Emacs _should_ do that.
I don't think we are (or will be) telling users to _not_ use the Customization UI either.  We cannot ignore the possibility that some user will use `customize` and if they do, the results should not be unexpected.

I believe you have the correct idea to _document_  the loading of `custom.el` should the user _want_ to make use of values set through the UI. I don't believe we are targeting users who have no idea how to add their own elisp configuration (the Principles state we are even encouraging it) , especially with an example of what to do.

jeastman (changes requested):
After some reflection on this, I think we need to add some documentation in the README to cover this change. It could result in confusing behavior if we don't make it clear what is going on. I think you have the right justification in your comments, we just need to make sure the user knows about it. Would you mind including the documentation in your PR?

jeffbowman:
I added some docs as requested, however, the PR does not appear to be updated...

jeastman (approved):
Nice job on the docs! Thanks.

jeastman:
Thank you for the contribution!

PR #85, opened 2022-02-23
Add module rational-persistence.el (reopen)

hey-jeldo:
Sorry, i am new to all this, renamed my master branch because i initially din't branch for adding the module. Reopen of #75 ...   :^|

Some settings related to persistence of emacs over different sessions.
I like these a lot and didn't know about them for the longest time.
Maybe this is a way to expose these settings for users that are interested in this behaviour.

### DIFF
--- a/modules/rational-persistence.el
+++ b/modules/rational-persistence.el
@@ -1,0 +1,35 @@
+;;; rational-persistence.el -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2022
+;; SPDX-License-Identifier: MIT
+
+;; Author: System Crafters Community
+
+;; Commentary
+
+;; Some settings to make emacs more of a continous expierience.
+
+;;; Code:
+
+;; When you visit a file, point goes to the last place it was when file was previously visited.
+(save-place-mode 1)
+
+;; Saving of history (minibuffer, kill-ring, search-ring, keyboard macros, shell commands).
+(setq savehist-save-minibuffer-history t)
+  (setq savehist-additional-variables
+        '(kill-ring
+          search-ring
+          regexp-search-ring
+          last-kbd-macro
+          kmacro-ring
+          shell-command-history))
+
+(savehist-mode)
+
+;; Restore last opened buffers after closing emacs.
+(setq desktop-restore-eager 5)
+(desktop-save-mode)
+
+(provide 'rational-persistence)
+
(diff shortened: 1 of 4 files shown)


ajxn:
You should probably remove commit e029 an de50f?

modules/rational-persistence.el
```diff
@@ -0,0 +1,35 @@
...
+
+;;; Code:
+
+;; When you visit a file, point goes to the last place it was when file was previously visited.
+(save-place-mode 1)
+
+;; Saving of history (minibuffer, kill-ring, search-ring, keyboard macros, shell commands).
+(setq savehist-save-minibuffer-history t)
+  (setq savehist-additional-variables
+        '(kill-ring
```
daviwil, line 20: It's probably better not to include the kill ring because clipboard copies might contain sensitive information.

modules/rational-persistence.el
```diff
@@ -0,0 +1,35 @@
...
+          regexp-search-ring
+          last-kbd-macro
+          kmacro-ring
+          shell-command-history))
+
+(savehist-mode)
+
+;; Restore last opened buffers after closing emacs.
+(setq desktop-restore-eager 5)
+(desktop-save-mode)
```
daviwil, line 31: I'm not as sure about turning this on by default.  Personally, I like having a clean Emacs session by default.  Anyone else have thoughts on this?

abarocio80: Some times my emacs state is so messedup that I need a fresh restart, and some of this configurations make that "fresh restart" a real headache. Maybe if we have some kill switch to delete this stateful behaviour?

andyhunne: As long as recent files are easily found in history, then I agree - I like to start with a clean session. Personal preference I guess.

modules/rational-persistence.el
jeastman: Maybe the whole value of `savehist-additional-variables` should be avoided. We could document it as something a user _should_ configure in their own configuration.

modules/rational-persistence.el
jeastman: Maybe this serves as a good custom configuration variable, something like `rational-persistence-restore-session` which sets and configures `desktop-save-mode` as well as anything else needed to resume where the user left off when relaunching emacs. This could be set to off (`nil`) by default.

PR #87, opened 2022-02-24
Load theme instead of setting default fg/bg color

jeffbowman:
Configuring the default foreground and background color takes extra
effort to undo when loading a theme after early-init processes. See
#37 for commentary and resolution.

Loads the deeper-blue theme as that one is built-in and matches
closely the default colors.

### DIFF
--- a/early-init.el
+++ b/early-init.el
@@ -26,9 +26,10 @@
 (push '(tool-bar-lines . 0) default-frame-alist)
 (push '(menu-bar-lines . 0) default-frame-alist)
 (push '(vertical-scroll-bars) default-frame-alist)
-(push '(background-color . "#232635") default-frame-alist)
-(push '(foreground-color . "#FFFFFF") default-frame-alist)
 (push '(mouse-color . "white") default-frame-alist)
+
+;; Loads a nice blue theme, avoids the white screen flash on startup.
+(load-theme 'deeper-blue t)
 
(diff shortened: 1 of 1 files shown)


early-init.el
```diff
@@ -26,10 +26,11 @@
 (push '(tool-bar-lines . 0) default-frame-alist)
 (push '(menu-bar-lines . 0) default-frame-alist)
 (push '(vertical-scroll-bars) default-frame-alist)
-(push '(background-color . "#232635") default-frame-alist)
-(push '(foreground-color . "#FFFFFF") default-frame-alist)
 (push '(mouse-color . "white") default-frame-alist)
 
+;; Loads a nice blue theme, avoids the white screen flash on startup.
+(load-theme 'deeper-blue t)
```
daviwil, line 32: I wonder how well this will work with Emacs running as a daemon?

ajxn: Maybe place a guard (test) there and only load when `display-graphic-p` and/or not `daemonp`?

jeffbowman: Works fine. I usually run emacs as a deamon and then use emacsclient to start emacs.

ajxn: Test it with `emacs -nw` and `emacs-client -c`  in a terminal emulator should to verify that it works.

Could probably use a customize variable to make it easy to change?

PR #90, opened 2022-02-26
Fix rational modules skeleton regex

### LINKED ISSUE #89: Skeleton autoinsert template is active on all `.el` files
Whenever I open a new `.el` file emacs offers to "Perform Rational Emacs Lisp Skeleton auto-insertion? (y or n)".

But as I understand the form for autoinsertion, it seems to be offering it only on files wich names starts with "rational-" and ends in ".el", not only the latter.

I still can choose to not accept the offering, but I wonder if it would interfer if some one wants his/hers own auto insertion forms.

Any way — Is this the intended behaviour for this skeleton, or is this a bug?

jeffbowman:
Expected the regex to only apply to the filename without the path,
however, the code uses `buffer-file-name` which includes the path. So
personal modules whose path matched the regex resulted in the module
skeleton insertion which was not intended. This should fix #89 and
only insert for elisp files found in the modules folder and named
rational-[ *module* ].el

### DIFF
--- a/init.el
+++ b/init.el
@@ -62,7 +62,8 @@ straight.el or Guix depending on the value of
 (auto-insert-mode)
 (with-eval-after-load "autoinsert"
   (define-auto-insert
-    (cons "rational-.*\\.el" "Rational Emacs Lisp Skeleton")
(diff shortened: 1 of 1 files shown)


daviwil:
Uh oh, thanks for the fix!

PR #92, opened 2022-02-28
added rational-modeline.el

### LINKED ISSUE #44: make doom-modeline optional

erikLundstedt:
this closes #44 and replaces #45 

moved existing configs to the new module
and added 2 lines to example file to show how to use it

Signed-off-by: erik.lundstedt <erik@lundstedt.it>


modules/rational-modeline.el
```diff
@@ -0,0 +1,134 @@
...
+
+;; Commentary
+
+;;
+
+;;; Code:
+
+(defcustom rational-modeline/type "doom"
+  "one of 'GNU' 'doom' 'telephone' ";;'space WIP' "
+  )
```
jeastman, line 16: You will need to handle the situation where this changes (e.g. with `:set`), otherwise it will not take effect unless it is configured _before_ the module is required.

`rational-modeline/type` should be `rational-modeline-type` for consistency with style guide.

modules/rational-modeline.el
```diff
@@ -0,0 +1,134 @@
...
+  ;; (defun test ()            ;;
+  ;;   (message "hello world") ;;
+  ;;   )                       ;;
+  ;;                           ;;
+  ;;   (setq a 'test)          ;;
+  ;;                           ;;
+  ;;                           ;;
+  ;;                           ;;
+  ;; (funcall a)               ;;
+  ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
```
jeastman, line 116: Probably should remove the test comment block

modules/rational-ui.el
```diff
@@ -13,7 +13,7 @@
 ;;; Code:
 
 (straight-use-package 'all-the-icons)
-(straight-use-package 'doom-modeline)
+;;(straight-use-package 'doom-modeline)
```
jeastman, line 16: No need to leave the comment in file here

modules/rational-ui.el
```diff
@@ -27,16 +27,18 @@ Use a plist with the same key names as accepted by `set-face-attribute'.")
...
+;;                                                                                         ;;
+;; ;; Start up the modeline after initialization is finished                               ;;
+;; (add-hook 'after-init-hook 'doom-modeline-init)                                         ;;
+;;                                                                                         ;;
+;; ;; Configure `doom-modeline'                                                            ;;
+;; (customize-set-variable 'doom-modeline-height 15)                                       ;;
+;; (customize-set-variable 'doom-modeline-bar-width 6)                                     ;;
+;; (customize-set-variable 'doom-modeline-minor-modes t)                                   ;;
+;; (customize-set-variable 'doom-modeline-buffer-file-name-style 'truncate-except-project) ;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
```
jeastman, line 41: No need to leave comment in file here.

modules/rational-modeline.el
```diff
@@ -0,0 +1,134 @@
...
+;;
+
+;;; Code:
+
+(defcustom rational-modeline/type "doom"
+  "one of 'GNU' 'doom' 'telephone' ";;'space WIP' "
+  )
+
+
+(when (string-equal rational-modeline/type "GNU")
```
jeastman, line 19: Each `when` would need to be handled when the custom variable changes.

modules/rational-modeline.el
```diff
@@ -0,0 +1,134 @@
...
+  ;;TODO make it look as good as possible without external packages
+  )
+
+
+
+
+(when (string-equal rational-modeline/type "doom")
```
jeastman, line 26: Each `when` would need to be handled when the custom variable changes.

modules/rational-modeline.el
```diff
@@ -0,0 +1,134 @@
...
+  (customize-set-variable 'doom-modeline-height 15)
+  (customize-set-variable 'doom-modeline-minor-modes t)
+  )
+
+
+
+
+
+
+(when (string-equal rational-modeline/type "telephone")
```
jeastman, line 40: Each `when` would need to be handled when the custom variable changes.

modules/rational-modeline.el
```diff
@@ -0,0 +1,134 @@
...
+    )
+
+
+
+
+
+  (straight-use-package 'telephone-line)
+
+
+  (defcustom rational-modeline/telephoneline-style 'telephoneline-examples-cubed
```
jeastman, line 101: Also want a `:set` here to handle when the value changes.

jeastman (changes requested):

modules/rational-modeline.el
erikLundstedt: > You will need to handle the situation where this changes (e.g. with :set)
not sure how to do this, happy to get an example or something like that.

I havent really written a lot of elisp code and allmost none outside of my personal configs where this doesnt matter too much

erikLundstedt: > rational-modeline/type should be rational-modeline-type for consistency with style guide.
this should be fixed now



not sure how to add changes to pr... but i fixed this in a recent push

modules/rational-modeline.el
erikLundstedt: How would I do this?

I am pretty new to elisp and would be glad for any help you can give

modules/rational-modeline.el
jeastman: > I havent really written a lot of elisp code and allmost none outside of my personal configs where this doesnt matter too much

No worries. One of the Principles of Rational Emacs: "Helps you learn Emacs Lisp"

modules/rational-modeline.el
jeastman: Check out the implementation of `rational-ui-display-line-numbers` in rational-ui.el. Notice the `:set` on the custom variable and how it responds to the variable being changed. Just make sure you use `set-default` in the body of your `:set` to ensure the value actually gets changed. You can test by setting the value (with `custom-set-variables`). Try setting the value both before and after `require`ing `rational-modeline` to see the effect.  Happy to help if you have questions after looking at `rational-ui-display-line-numbers`.

modules/rational-modeline.el
```diff
@@ -0,0 +1,158 @@
...
+
+;;; Code:
+
+(defcustom test 'nil
+  "something here"
+  :group 'rational
+  :set (lambda (sym val)
+         (set-default sym val)
+         (when (equal test t)
+           (counsel-find-file)
```
bkaestner, line 20: This shouldn't be in a productive pull request. Please remove the test code.

modules/rational-modeline.el
```diff
@@ -0,0 +1,158 @@
...
+
+;;(when (string-equal rational-modeline-type "GNU")
+;;TODO make it look as good as possible without external packages
+;;  )
+
+
+
+
+(defun doom-modeline-type ()
+  (straight-use-package 'doom-modeline)
```
bkaestner, line 34: None of these functions have either the `rational-` prefix, nor a docstring. Please add some documentation.

modules/rational-modeline.el
```diff
@@ -0,0 +1,158 @@
+;;;; rational-modeline.el --- rational module for setting up the modeline  -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2022
+;; SPDX-License-Identifier: MIT
+
+;; Authors: System Crafters Community, David "System Crafters" Wilson, Erik Lundstedt
```
bkaestner, line 6: I don't think I have heard of @daviwil call himself _David "System Crafters" Wilson_ 🤔

custom-vars.el
```diff
@@ -0,0 +1,39 @@
...
+ '(scroll-conservatively 101)
+ '(scroll-margin 0)
+ '(scroll-preserve-screen-position t)
+ '(visible-bell 1))
+(custom-set-faces
+ ;; custom-set-faces was added by Custom.
+ ;; If you edit it by hand, you could mess it up, so be careful.
+ ;; Your init file should contain only one such instance.
+ ;; If there is more than one, they won't work right.
+ )
```
bkaestner, line 39: This whole file seems like an accidental commit. Was this really intended? Otherwise please rebase to remove the file.

bkaestner (changes requested):

custom-vars.el
erikLundstedt: Yeah that was accidental

erikLundstedt:
I am thinking that if the user wants to use something not supported by the module, they can just not require it, at the moment I'm using this branch+with the setting for telephone-line which works great for me and seems to be pretty much what you wanted (according to what you(@daviwil ) said about it in the video) by default the module uses doom modeline though as I don't want to break configs, I also added the example configs for telephone-line for ease of use but the user is able to write their own if they want

Telephone-line seems to do for the modeline what vertico does for completion, it makes it look good while keeping it simple, lightweight and vanilla-like

custom-vars.el
jeastman: You will need to remove the file from your PR as it does not belong.

history
```diff
@@ -0,0 +1,16 @@
+;; -*- mode: emacs-lisp; coding: utf-8-unix -*-
```
jeastman, line 1: This file also does not appear to be related to the commit. Another accidental inclusion?

modules/rational-modeline.el
jeastman: Yes, should probably update the comment.

modules/rational-modeline.el
jeastman: Agree this should also be removed.

jeastman (changes requested):

modules/rational-modeline.el
jeffbowman: For the choice of options, you should probably use something like the following in your `defcusom` definition:

``` emacs-lisp
    :type '(choice (const :tag "GNU default" gnu)
                   (const :tag "Doom modeline" doom)
                   (const :tag "Telephone modeline" telephone))
```

erikLundstedt:
im closing this as i moved the code to my own config folder

PR #93, opened 2022-03-04
Add rational defaults for ibuffer and whitespace

jeffbowman:
Use ibuffer as the buffer list, configure it to group by project.
Use whitespace to cleanup whitespace in a buffer.

### DIFF
--- a/modules/rational-defaults.el
+++ b/modules/rational-defaults.el
@@ -11,6 +11,8 @@
 
 ;;; Code:
 
+(straight-use-package 'ibuffer-project)
+
 ;; Revert Dired and other buffers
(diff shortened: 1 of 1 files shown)


bkaestner (changes requested):
Hm. This PR does three distinct things:

- it adds `ibuffer-project` as an external dependency and uses it as an alternative to `list-buffers`
- it configures `whitespace-mode`
- it fixes a byte-compilation warning

I *personally* would like three PRs instead, and *definitely* two distinct commits for the `ibuffer-project` and `whitespace` changes.

daviwil:
I'm not sure that I'd prefer `ibuffer` to be the default buffer switcher, but I'll think on this one for a bit.

PR #100, opened 2022-03-17
Add whitespace configuration

jeffbowman:
After feedback on #93, split whitespace configuration to its own PR. 

`whitespace.el` is built-in to Emacs, provides visualization of whitespace at ends of lines or where tabs are used, etc. This configuration turns on those features as well as configures whitespace.el to cleanup whitespace on save.


jeffbowman:
This is a bit of a mess... closing and will reopen in a bit when I've cleaned up the git history.

PR #101, opened 2022-03-17
Remove ws-butler in favor of whitespace.el

jeffbowman:
whitespace.el provides what ws-butler was doing and is built
in. Following the principles, preferring whitespace.el.

### DIFF
--- a/modules/rational-editing.el
+++ b/modules/rational-editing.el
@@ -11,15 +11,17 @@
 
 ;;; Code:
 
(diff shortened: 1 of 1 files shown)


jeastman (approved):
Looks good. 

Only suggestion for improvement would be to make the modes for which `whitespace-mode` is enabled be configurable, but that would be an enhancement.

JulienTant:
FYI, it's pretty ugly by default for tabs lol
image

PR #104, opened 2022-03-21
Fix setting for eln-cache location -- closes #103

### LINKED ISSUE #103: set eln-cache depending on emacs version
Right now the setting of eln-cache is

```emacs-lisp
 ;; Set the right directory to store the native compilation cache
  (add-to-list 'native-comp-eln-load-path (expand-file-name "eln-cache/" user-emacs-directory)))
```
This is the default. But should the user decide to change the location it won't work properly on emacs 29. Instead  you need to use the new function `startup-redirect-eln-cache`. You can see the report from this [bug report](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=53891) and the related [stack exchange discussion](https://emacs.stackexchange.com/a/70478/11934). So I recommend putting something like the following instead:

```emacs-lisp
;; Put eln-cache dir in cache directory
;; NOTE the method for setting the eln-cache dir depends on the emacs version
(when (fboundp 'startup-redirect-eln-cache)
  (if (version< emacs-version "29")
      (add-to-list 'native-comp-eln-load-path (concat "path/to/etc/dir" "eln-cache/"))
    (startup-redirect-eln-cache (concat "path/to/etc/dir" "eln-cache/"))))
```

mclearc:
- Put eln-cache in var folder (NOTE: had to move config path settings to make
this work)
- Set conditional for moving eln-cache based on emacs version

### DIFF
--- a/early-init.el
+++ b/early-init.el
@@ -7,33 +7,6 @@
 ;; Prefer loading newest compiled .el file
 (setq load-prefer-newer noninteractive)
 
-;; Native compilation settings
-(when (featurep 'native-compile)
-  ;; Silence compiler warnings as they can be pretty disruptive
-  (setq native-comp-async-report-warnings-errors nil)
-
-  ;; Make native compilation happens asynchronously
-  (setq native-comp-deferred-compilation t)
-
-  ;; Set the right directory to store the native compilation cache
-  (add-to-list 'native-comp-eln-load-path (expand-file-name "eln-cache/" user-emacs-directory)))
-
-;; Don't use package.el, we'll use straight.el instead
-(setq package-enable-at-startup nil)
-
-;; Remove some unneeded UI elements (the user can turn back on anything they wish)
-(setq inhibit-startup-message t)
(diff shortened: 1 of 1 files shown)


early-init.el
```diff
@@ -55,7 +28,7 @@
    ((getenv "RATIONAL_EMACS_HOME") (expand-file-name (getenv "RATIONAL_EMACS_HOME")))
    ((or (getenv "XDG_CONFIG_HOME") (file-exists-p (expand-file-name ".config/rational-emacs" (getenv "HOME"))))
     (if (getenv "XDG_CONFIG_HOME")
-	(expand-file-name "rational-emacs" (getenv "XDG_CONFIG_HOME"))
+	    (expand-file-name "rational-emacs" (getenv "XDG_CONFIG_HOME"))
```
jeffbowman, line 31: minor nit... looks like indentation got a little extra space here.

jeffbowman (approved):
Looks good, thanks!

PR #120, opened 2022-04-04
Migrate to package.el

### LINKED ISSUE #72: Why have rational-use-package?
Just wondering why have the rational-use-package module? Given the decision to use straight.el rather than package.el, I don't understand why we also have rational-use-package.el, which will use package.el to install packages instead of just doing 

```
(straight-use-package 'use-package)
(setq straight-use-package-by-default t)

```

which enables the use of use-package, but without the addition of a second package management framework. Is there something I'm missing here?

### LINKED ISSUE #94: straight.el vs package.el - Which should we stick with?
Discuss!

jeffbowman:
Replaces usage of `straight-use-package` with `package-install`, but uses a macro to facilitate that. I tried to keep things reversible by checking to make sure `package-enable-at-startup` is `t` to allow for someone to bootstrap `straight.el` in their personal configuration. 

Fixes #94 
Possibly closes #72 as well


init.el
```diff
@@ -13,6 +13,54 @@
...
+   (dolist (archive package-archives result)
+     (let* ((archive-name (expand-file-name (format "archives/%s/archive-contents" archive)
+                                            package-user-dir))
+            (archive-modified-date (time-to-days
+                                    (file-attribute-modification-time
+                                     (file-attributes archive-name)))))
+       (when (time-less-p archive-modified-date today)
+         (setq result t))))))
+
+(defmacro rational-install-package (package)
```
jeastman, line 33: Should this be `rational-package-install-package`? Seems a bit redundant, but would match the naming conventions.

modules/rational-use-package.el
```diff
@@ -12,22 +12,26 @@
...
-
-;; Always add melpa to `package-archives'
-(add-to-list 'package-archives (cons "melpa" "https://melpa.org/packages/"))
-
-(package-initialize)
-(unless package-archive-contents
-  (package-refresh-contents))
+(rational-install-package 'use-package)
+
+;; we are possibly here because `straight.el' is preferred, so only
```
jeastman, line 17: Again confused by the reference to straight.

init.el
```diff
@@ -13,6 +13,54 @@
...
+  "Use `package.el' when it is enabled. Switch to using
+`straight.el' if it is not."
+  (if package-enable-at-startup
+      `(unless (package-installed-p ,package) (package-install ,package))
+    `(straight-use-package ,package)))
```
jeastman, line 38: While falling back to straight could be an interesting idea, straight will not work. The bootstrap code for straight is removed below. Is the expectation here that straight could still be used if the user decides to set it up on their own?

jeastman (approved):
Despite being confused by the role of straight here, the changes look fine to me.  I'll work to sort out my ignorance in #94

init.el
jeffbowman: I can fix that... but not necessary I get rid of the macro. (see #94 response)

init.el
jeffbowman: I'd rather get rid of it. Initially I thought it would be good to have something like this, but I think I agree that it confuses things. I'll update the PR and remove it.

jeastman (approved):
Looks good. Thanks!

johnhamelink:
Unfortunately I've been caught out attempting to continue to use `straight-use-package` after this update - I've created my own deviation of the `rational-ui` file to revert back to using straight for the installation of `doom-modeline` as a stop-gap solution to get my config back up and running.

jeffbowman:
@johnhamelink, I have added a minimal example config in the `examples/straight` folder to demonstrate using `straight.el` with your config if that helps you.

johnhamelink:
Thank you, that did help! @jeffbowman

PR #122, opened 2022-04-06
activate org-appear when org-mode is activated

ingorichter:
Following the installation and usage instructions of [org-appear](https://github.com/awth13/org-appear) it seems that org-appear-mode was added to the wrong hook and didn't get activated


jeffbowman (approved):
Nice catch! Thanks!

Issue #64, opened 2022-02-19
rational ide module


HumHongeKamyaab:
I think the prime purpose of emacs is to use it as a modern IDE. It would greatly help the community if this project could provide a minimal module like a framework for different programming languages by integrating [eglot](https://github.com/joaotavora/eglot) (or [lsp](https://github.com/emacs-lsp/lsp-mode)), [yaspinnet](https://github.com/joaotavora/yasnippet), [company mode](https://github.com/company-mode/company-mode), and flymake. 

[Doom Emacs](https://github.com/hlissner/doom-emacs) and [Spacemacs](https://github.com/syl20bnr/spacemacs) already try to do it using layering. However, it is still confusing, and snippets from these projects can't be integrated into user configs easily, which is the best part of the rational emacs project.

jeastman:
Before we touch on the specific packages to integrate (`company mode` could be replaced with [corfu](https://github.com/minad/corfu) which could be argued to be more closely aligned with the Principles), let's define the functionality one would expect from a "modern IDE".

I am no expert when it comes to "modern IDE"s, but I suspect the beginning of that list would be:

- In-line completion while typing in source code
- Support for templates / abbreviation expansion
- Automatic verification / checking / compilation
- Automatic formatting to code standard
- Code overview tools (source outline, etc.)
- Complex refactoring support

The next decision is if these go into a single module, like `rational-ide` or are the specific features more generally applicable? Maybe we could try to align each specific feature into existing module as a first step and, failing that, decide on a more focused `rational-ide` module.

jeastman:
The more I think about this, the more I think we need to answer the second question first... Should we have a `rational-ide` module?

theophilusx:
I think it would be very useful to have a rational-ide module, but with a 'rational' twist. 

Emacs has amazing functionality built-in which many users don't even realise. There are lots of features/modes built-in which many users never even realise and many of the external packages which people add actually re-implement features which are already there (but often with a slightly different approach or which use slightly different terminology which makes them difficult to recognise. 

What I would like to see is a rational-ide which tries to create a modern IDE using just built-in modes/packages from Emacs and GNU ELPA (and maybe NONGNU ELPA). Where we do find it necessary to go with 'external' packages, prefer ones which also attempt to leverage as much as possible on built-in facilities. For example, eglot over lsp-mode, flymake over flycheck, project.el over projectile etc. 

[...]

PR #123, opened 2022-04-12
config for rational-ide

HumHongeKamyaab:
little step to resolve https://github.com/SystemCrafters/rational-emacs/issues/64

### DIFF
--- a/modules/rational-ide.el
+++ b/modules/rational-ide.el
@@ -1,0 +1,32 @@
+;; rational-ide.el -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2022
+;; SPDX-License-Identifier: MIT
+
+;; Author: System Crafters Community
+
+;; Commentary
+
+;; Eglot configuration
+
+;;; Code:
+
+
+;; Install dependencies
+(rational-package-install-package 'xref)
+(rational-package-install-package 'project)
+(rational-package-install-package 'eldoc)
+(rational-package-install-package 'yasnippet)
+(rational-package-install-package 'yasnippet-snippets)
+
+
+
+(rational-package-install-package 'eglot)
+
+
+;; hooks
+(add-hook 'prog-mode-hook 'eglot-ensure)
+(add-hook 'prog-mode-hook 'yas-minor-mode)
+
+(add-hook 'LaTeX-mode-hook 'eglot-ensure)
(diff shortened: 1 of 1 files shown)


abarocio80:
Looks good.  But I don't think of LaTeX as part of an IDE, even when it is a useful tool.  Wouldn't that be better on another module?  Just a thought.

jeffbowman:
Good start, I think the lisp family of languages usually have something already (although there are LSPs for Clojure and Racket it seems) ala `geiser` for scheme languages, `sly` or `slime` for common lisp, `cider` for Clojure. So, not sure we would want to run `eglot-ensure` in those cases. Maybe instead of adding it to `prog-mode`, we could loop over `eglot-server-programs` as it mentions the major modes supported.

Something along the lines of (untested, might not work):

```emacs-lisp
(defun rational-ide--eglot-add-hook (mode)
  (let ((hook-name (intern (concat (symbol-name mode) "-hook"))))
    (add-hook hook-name #'eglot-ensure)))

(mapcar #'rational-ide--eglot-add-hook eglot-server-programs)
```

bdarcus:
> ... I don't think of LaTeX as part of an IDE, even when it is a useful tool. Wouldn't that be better on another module?

It definitely belongs in its own module; `rational-latex`.

jeffbowman:
> It definitely belongs in its own module; rational-latex.

Maybe something more generic like `rational-writing` would be more appropriate. Such a module might include some configuration for blog writing, maybe olivetti, LaTeX, etc.

bdarcus:
> Maybe something more generic like `rational-writing` would be more appropriate. Such a module might include some configuration for blog writing, maybe olivetti, LaTeX, etc.

True, but LaTeX is only one writing mode, and there's already an org module, which is arguably more widely used in the emacs world for writing.

:shrug: 

Perhaps they could be complementary, so one could couple LaTeX/markdown/org with such a mode.

jeffbowman:
> True, but LaTeX is only one writing mode, and there's already an org module, which is arguably more widely used in the emacs world for writing.

Fair point. No argument from me :-)

> Perhaps they could be complementary, so one could couple LaTeX/markdown/org with such a mode.

Ah, markdown mode, I had forgotten that one. But, yes, I like this idea. 

(in the interest of not polluting the discussion for this PR, maybe these ideas should go into their own issue...)

PR #129, opened 2022-04-20
Disable auto-pairing of "<" in org-mode

sthesing:
With the example config, electric-pair-mode is active in org-mode and it pairs all brackets, including "<>".
While that *is* useful for e.g. manually entering agenda dates, I propose  to disable it by default, because

a) most people probably use functions like `org-schedule` for those
b) many useful [snippets](https://github.com/AndreaCrotti/yasnippet-snippets/tree/master/snippets/org-mode) from the widely used yasnippet package begin with "<".

I've taken the liberty to prepare a PR disabling it in the module `rational-completion`, but I'm unsure whether that's the best place for this bit of code. If you want it, at all, that is.

### DIFF
--- a/modules/rational-completion.el
+++ b/modules/rational-completion.el
@@ -76,5 +76,11 @@ folder, otherwise delete a word"
 ;; Use Embark to show bindings in a key prefix with `C-h`
 (setq prefix-help-command #'embark-prefix-help-command)
 
+;; disable auto-pairing of "<" in org-mode
+(add-hook 'org-mode-hook (lambda ()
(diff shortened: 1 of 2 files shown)


jeffbowman:
As this seems to apply to `org-mode`, I'd suggest moving this to the `rational-org` module instead.

sthesing:
> As this seems to apply to `org-mode`, I'd suggest moving this to the `rational-org` module instead.

Good suggestion, thanks for the feedback. PR updated.

jeffbowman:
Thanks for the contribution!!

PR #131, opened 2022-04-24
Change theme in example-config.el to `doom-one`

sthesing:
In its current state, starting a fresh install of rational emacs with the example config fails. 
One reason is that it tries to load `doom-snazzy`, which isn't available the last versioned release of doom-themes.

`doom-one` has a similar look and is available.

### DIFF
--- a/README.org
+++ b/README.org
@@ -192,7 +192,7 @@ filesystem, it is created. However, just the path is created, the files
    '(rational-ui-default-font
      '(:font "JetBrains Mono" :weight light :height 185)))
 
-  (load-theme 'doom-snazzy t)
+  (load-theme 'doom-one t)
 
   ;; To not load `custom.el' after `config.el', uncomment this line.
   ;; (setq rational-load-custom-file nil)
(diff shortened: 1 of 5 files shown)


jeffbowman:
Looks good. Unfortunately, there are 2 instances of this file, the one you changed and a duplicate in `examples/general/example-config.el` which needs the same fix. Would you mind adding that to your PR?

sthesing:
Oh, I overlooked that. Thanks for pointing it out. I'll amend the PR.
I grepped through the repo and found more mentions of `doom-snazzy` in the `docs` directory. Does it make sense to change these, too? Or are the generated from the example configs, anyway?

jeffbowman:
It makes sense to change those as well. Thanks for finding them!

sthesing:
> It makes sense to change those as well. Thanks for finding them!

OK, thanks for getting back to me with that. Changed all of them. Including README

jeffbowman:
Thanks very much!

PR #135, opened 2022-04-26
Remove evil-nerd-commenter

jeffbowman:
Using comment-dwim is built-in and closer to the principles of the
project.

The built-in comment-dwim functionality does the same thing with a
couple of caveats:

* comment-dwim needs a region to comment, evil-nerd-commenter allows
also specifying the number of lines from "here" to comment
* comment-dwim allows end-of-line comments, evil-nerd-commenter does
not
* comment-dwim has a method to remove comments that
evil-nerd-commenter lacks.

### DIFF
--- a/docs/rational-editing.org
+++ b/docs/rational-editing.org
@@ -6,19 +6,6 @@
     (require 'rational-editing)
   #+end_src
 
-  - =evil-nerd-commenter=
-
-    This module installs the =evil-nerd-commenter= package to provide
-    a few more features when commenting code; for example, you can
-    provide a number of lines to comment and press a key to do the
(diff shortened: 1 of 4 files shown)


sthesing:
Good call, imho.
Not being an evil user, I wonder: does evil-nerd-commenter do anything in a vimmy way that comment-dwim doesn't? If so, maybe it could stay semi-default and get loaded by default if users turn on `rational-evil`.

jeffbowman:
Yes, it does. `C-u 5 M-x evilnc-comment-or-uncomment-lines` will comment the next 5 lines including this one. 

Not a bad idea, I could move it to the `rational-evil` module, that probably makes sense.

PR #136, opened 2022-04-26
Enable loading modules provided by the user.

### LINKED ISSUE #29: make it read ~/.config/rational-emacs/custom-modules for...
`dramatic pause`
...custom modules, perfect for modularizing your own config without having to follow the [principles](https://github.com/SystemCrafters/rational-emacs/blob/master/README.org#principles) (this might fit better in something like personal-modules or equivalent) and for keeping a personal or public list of modules that cant/havent yet been merged into the regular modules directory

@daviwil

telenieko:
If there is a `custom-modules` directory under `rational-config-path`
it gets added to `load-path` so the user can put there their own modules.

Also, it is added at the top of the `load-path` so it would be possible
for a user to override another module (for example, provide their own `rational-ui.el`)

Fixes #29

### DIFF
--- a/init.el
+++ b/init.el
@@ -57,6 +57,18 @@ Return non-nil if the on-disk cache is older than one day or
 ;; Add the modules folder to the load path
 (add-to-list 'load-path (expand-file-name "modules/" user-emacs-directory))
 
+;; Add the user's custom-modules to the top of the load-path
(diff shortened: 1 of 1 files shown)


jeffbowman:
Nice addition, thanks!

PR #137, opened 2022-04-26
Ignore `bookmarks` file which shall never be commited

telenieko:
Emacs default bookmarks file is within `user-emacs-directory`.
So maybe add it to `.gitignore`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ eshell/
 rational-emacs/
 projects
 elpa/
+bookmarks


jeffbowman:
Good idea! Maybe configure it to be in the `rational-config-path`/etc folder as part of this PR?

telenieko:
> Good idea! Maybe configure it to be in the `rational-config-path`/etc folder as part of this PR?

I did want to go there as that would be a breaking change for any current user, the intent here now is to avoid anyone from accidentally pushing their `bookmarks` file on a PR or having a permanently *dirty* directory for that, and probably other, files.

Maybe there should be a separate issue titled "Have user-data live in `rational-config-path`" so a discussion can be had on what files that normally reside in `user-emacs-directory` could be moved to `rational-config-path`? 

If ok I'll leave this one as-is and create that other issue before going deeper.

jeffbowman:
I think just adding it to the .gitignore file is not sufficient, also need to do something like `git rm --cached bookmarks` for git to completely forget about the file and ignore it in the future.

We do have the disclaimer a the top of the README.org file that things are still in development, so a breaking change would be at least not unexpected. But I see your point, we can defer.

PR #138, opened 2022-04-26
Move all theme related stuff to the Example

### LINKED ISSUE #133: Make the theme customizable / optional
The current coding of `rational-ui` installs `doom-themes` always.

This should be optional as the user may want to use other themes.

Any pointers on how that could be achieved?

telenieko:
See discussion in #133.
Theme related stuff is now only in the example configuration with a note to make it clear to the user they should load some theme.

Closes #133.


jeffbowman:
Thanks!!

PR #139, opened 2022-04-27
Add doom-themes installation to examples and docs

sthesing:
Since #138 the loading of doom-one requires installation of
`doom-themes` by the user. It's only present in one example. This
adds it to all examples and docs.

So - depending on which example config file the user starts out with - a new user might end up with an error instead of a successful startup of Emacs.

Alternatively, we could remove the loading of `doom-one` from all examples and docs.


jeffbowman:
Thanks! Probably need to consolidate those example configurations to just the `examples` folder.

PR #141, opened 2022-04-27
Run package-contents-refresh in the background for stale archives

telenieko:
Though it makes sense to do it synchronously when the archive is empty,
the refresh due to stale archives could be done in the background thus greatly improving startup time.

Otherwise your first start every morning can take a while.


jeffbowman:
Nice! Thanks for the contribution!

Issue #144, opened 2022-05-03
rational-compile: only compile required modules


sthesing:
If a user requires the `rational-compile` module, it attempts to compile all modules, regardless whether the user uses them or not. That leads to compiler errors, because dependencies of unused modules probably aren't installed.

For example, even if a user doesn't require `rational-evil` in their `config.el`, `rational-compile` will attempt to compile `modules/rational-evil.el`. This will fail if `evil` isn't installed.

So it probably would be best to only compile the modules the user actually requires in their `config.el`.

(Tagging @abarocio80 the original author of `rational-compile`).

abarocio80:
That's a real issue.

I guess it is not that difficult to fix with a list of applied modules, but that would add to the user the task to maintain such a list.

I think that one way to address that problem would be to create a function (or macro) to apply the module and to add that module to the list of modules to compile.

Which way could be better? Or are there another approaches to this issue? I ask to the community.

jeffbowman:
Closing this after merging #145

PR #145, opened 2022-05-04
Proposal to fix issue #144

abarocio80:
This is a proposal to fix issue #144.

It will require to the user to mantain the list of modules to compile using the variable `rational-compile-module-list'.

If a tool (function or macro) will be developed to keep track of both requires and the mantainance of the module list to be compiled, it will be on another module.

I also made some minor modifications that had sense no more with this update, but I tried to keep the interface intact.

### DIFF
--- a/modules/rational-compile.el
+++ b/modules/rational-compile.el
@@ -35,7 +35,7 @@
   "A list of modules to compile.")
 
 (defvar rational-compile-modules-path
-  `(,(expand-file-name "modules" rational-config-path)
+  `(,(expand-file-name "custom-modules" rational-config-path)
     ,(expand-file-name "modules" user-emacs-directory))
   "Path where to locate modules.")
 
@@ -49,7 +49,10 @@
   "When non-nil, compile the file after saving it.")
 
 (defvar rational-compile-extra-directories-list nil
-  "List of extra directories to compile.")
+  "List of extra directories to search for source-files.
+
+This could be another directory added to `load-path', that is not
+standard, nor is it part of rational-emacs.")
 
 (defvar rational-compile-init-files-list
   '("early-init.el" "init.el")
@@ -63,16 +66,28 @@
 ;;;; Functions:
 ;; A function to locate a module's source file:
 (defun rational-locate-module-file (module &optional path)
-  (setq-local module-path nil)
-  (dolist (dir (or path
-                   (flatten-list `(,rational-compile-modules-path
-                                   ,rational-compile-extra-directories-list))))
-    (let ((file (expand-file-name (format "%s.el" (symbol-name module)) dir)))
-      (when (and (file-exists-p file)
-                 (not module-path))
-        (setq-local module-path file))))
-  module-path)
+  "Locates the source-file for the MODULE.
 
+MODULE must be a symbol.
+
+If the optional argument PATH is non-nil, it asumes it is a list
+containing the name of the directories to search for the
+source-file.  If it is ommited or nil, it will search in the
+directories defined in `rational-compile-modules-path' and
+`rational-compile-extra-directories-list'.
+
+It returns the first source-file that matches for the MODULE in
+the order specified search path.  If it finds none, it retorns
+nil."
+  (let ((file (expand-file-name
+               (format "%s.el" (symbol-name module))
+               (seq-find (lambda (dir)
+                           (file-exists-p (expand-file-name (format "%s.el" (symbol-name module)) dir)))
+                         (or path
+                             (flatten-list `(,rational-compile-modules-path
+                                             ,rational-compile-extra-directories-list)))))))
+      (when (file-exists-p file)
+        file)))
 
 ;; A function to compile a specific file
 (defun rational-compile-file (f)
@@ -120,7 +135,7 @@ D could be a single directory or a list of directories."
   (flatten-list `(,(and rational-compile-modules
                         (expand-file-name "modules/" user-emacs-directory))
                   ,(and rational-compile-user-modules
-                        (expand-file-name "modules/" rational-config-path))
+                        (expand-file-name "custom-modules/" rational-config-path))
                   ,rational-compile-extra-directories-list)))
 
 ;; A function to get the list of init files with full path:
@@ -179,7 +194,7 @@ specified, then it is ignored without a warning."
   (when rational-compile-user-modules
     (dolist (module rational-compile-module-list)
       (let ((module-src (rational-locate-module-file module
-                                                     `(,(expand-file-name "modules/" rational-config-path)))))
+                                                     `(,(expand-file-name "custom-modules/" rational-config-path)))))
         (when module-src
           (rational-compile-file module-src))))))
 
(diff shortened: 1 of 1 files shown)


modules/rational-compile.el
```diff
@@ -21,28 +21,59 @@
...
+(defvar rational-compile-config-files-list
+  '("early-config.el" "config.el")
+  "List of configuration file names.")
+
+

 ;;;; Functions:
+;; A function to locate a module's source file:
+(defun rational-locate-module-file (module &optional path)
+  (setq-local module-path nil)
```
jeffbowman, line 66: maybe this could be simplified to just:
```emacs-lisp
(seq-find (lambda (dir) (file-exists-p (expand-file-name (format "%s.el" (symbol-name module)) dir)))
          (or path (rational-compile--config-dirs-list)))
```

modules/rational-compile.el
```diff
@@ -21,28 +21,59 @@
...
+  "When non-nil, compile emacs lisp sources for the modules on startup.")
+
+;; We define a variable to enable/disable user's modules compilation.
+(defvar rational-compile-user-modules t
+  "When non-nil, compile user's emacs lisp sources for the modules on startup.")
+
+(defvar rational-compile-module-list '()
+  "A list of modules to compile.")
+
+(defvar rational-compile-modules-path
```
jeffbowman, line 37: This var may not be needed if the suggested change to the locate modules code is adopted (ie use the call to `rational-compile--config-dirs-list` instead).

modules/rational-compile.el
```diff
@@ -88,8 +119,8 @@ D could be a single directory or a list of directories."
   "Returns a list of configured directories to autocompile."
   (flatten-list `(,(and rational-compile-modules
                         (expand-file-name "modules/" user-emacs-directory))
-                  ,(and rational-compile-user-configuration
-                        (expand-file-name "./" rational-config-path))
+                  ,(and rational-compile-user-modules
+                        (expand-file-name "modules/" rational-config-path))
```
jeffbowman, line 123: After #136, this should be `custom-modules`. If we prefer just modules, will need an update to `init.el` as well to handle that name change.

modules/rational-compile.el
```diff
@@ -99,12 +130,58 @@ D could be a single directory or a list of directories."
...
+      (let ((module-src (rational-locate-module-file module
+                                                     (flatten-list
+                                                      `(,(expand-file-name "modules/" user-emacs-directory)
+                                                        ,rational-compile-extra-directories-list))))
+        (when module-src
+          (rational-compile-file module-src))))))
+  (when rational-compile-user-modules
+    (dolist (module rational-compile-module-list)
+      (let ((module-src (rational-locate-module-file module
+                                                     `(,(expand-file-name "modules/" rational-config-path)))))
```
jeffbowman, line 182: After #136, this should be `custom-modules`. If we prefer just modules, will need an update to `init.el` as well to handle that name change.

jeffbowman (commented):
Just a few suggestions, good work!

modules/rational-compile.el
abarocio80: This variable must be used, since is the way the user will define which modules will be compiled. If we only compile all the sources in the directories that are in the path, the issue #144 will not be solved.

The issue #144 is that this module will try to compile **all** the modules, even the ones not in use, which will cause dependency problems if all the dependencies are not installed.

This variable ensures that only the modules that the user want to compile will be compiled.

modules/rational-compile.el
abarocio80: That's great! I didn't remember about this function (`seq-find`); haven't heard of it in years and I totally forgot it.

modules/rational-compile.el
```diff
@@ -21,28 +21,59 @@
...
+
+;; We define a variable to enable/disable user's modules compilation.
+(defvar rational-compile-user-modules t
+  "When non-nil, compile user's emacs lisp sources for the modules on startup.")
+
+(defvar rational-compile-module-list '()
+  "A list of modules to compile.")
+
+(defvar rational-compile-modules-path
+  `(,(expand-file-name "modules" rational-config-path)
```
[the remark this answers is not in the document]
abarocio80: I wasn't aware of this.

abarocio80:
Sorry about that. Got some issues with my branches and I could not fix them in other way (need to learn to use git in a more proficient way).

modules/rational-compile.el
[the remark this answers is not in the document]
jeffbowman: No worries. :-)

modules/rational-compile.el
jeffbowman: Ah, I guess I misunderstood the code then, it appeared this was the path(s) to the modules and was only used in 2 places. Thanks for the clarification.

modules/rational-compile.el
```diff
@@ -63,16 +66,28 @@
...
+the order specified search path.  If it finds none, it retorns
+nil."
+  (let ((file (expand-file-name
+               (format "%s.el" (symbol-name module))
+               (seq-find (lambda (dir)
+                           (file-exists-p (expand-file-name (format "%s.el" (symbol-name module)) dir)))
+                         (or path
+                             (flatten-list `(,rational-compile-modules-path
+                                             ,rational-compile-extra-directories-list)))))))
+      (when (file-exists-p file)
```
jeffbowman, line 89: The `when` here is probably not needed (duplicates the check in the `lambda`). `file` will either be a fully qualified file name, or it will be `nil` based on the lambda passed to `seq-find`.

abarocio80: Yes! I didn't see that.

abarocio80:
Done!

abarocio80:
I realized the `let` was unnecessary in `rational-locate-module-file`, so I eliminated it.

abarocio80:
After all, the let was necessary and also the when, but I implemented those in another way.

modules/rational-compile.el
```diff
@@ -115,6 +205,16 @@ The files to be compiled is defined in
...
   (rational-compile-file (rational-compile--init-files-list)))
 
+;; A funciton to execute the compilation of the config files:
+(defun rational-compile-config ()
+  "Compile (native or byte-code) the configuration files.
+
+The files to be compiled is defined in
+`rational-compile-config-files-list'."
+  (interactive)
+  (rational-compile-file (rational-compile--init-config-list)))
```
jeffbowman, line 215: There is an error here: 

```emacs-lisp
Symbol's function definition is void: `rational-compile--init-config-list` 
```

jeffbowman (commented):
Added rational compile to my configuration and restarted Emacs which generated an error. 

Here is what I added to my config.el:

```emacs-lisp
(require 'rational-compile)
(with-eval-after-load "rational-compile"
   (add-to-list 'rational-compile-module-list 'rational-ui)
   (add-to-list 'rational-compile-module-list 'rational-defaults)
   (add-to-list 'rational-compile-module-list 'rational-completion)
   (add-to-list 'rational-compile-module-list 'rational-compile))
```

modules/rational-compile.el
jeffbowman: I fixed this.

PR #147, opened 2022-05-08
Fix example readme

### LINKED ISSUE #146: saving config.org does not create config.el
Hello
I copied example-configl.org to ~/.rational-emacs/config.org
I modified config.org and saving does not create config.el.
I did cp config.org config.el
changes to config.org are not showing in config.el once saved

sthesing:
I opted to write Option 1 mentioned in the discussion of #146 , simply because it involves the least necessary steps and thus is probably clearer.

This resolves #146.


jeffbowman (approved):
Thanks for the update!

Issue #128, opened 2022-04-19
Documentation for modules


jeffbowman:
I have created a `docs/` folder with an org file and a generated info file for providing documentation. 

Need help with documenting modules, see the `docs/CONTRIBUTING.org`. If you want to help, a PR is greatly appreciated! 

### craftedv2beta
Updated to the current focus (craftedv2beta)

- [x] crafted-completion (PR #331)
- [x] crafted-defaults (PR #345)
- [x] crafted-emacs-home (removed as of commit 5249c30)
- [x] crafted-evil (#344)
- [x] crafted-ide (PR #339)
- [x] crafted-lisp (PR #296)
- [x] crafted-org (PR #285)
- [x] crafted-osx (commit 928053072c1b434cc203829e3d44c69c213e02bc)
- [x] crafted-screencast (PR #303)
- [x] crafted-speedbar (PR #351)
- [x] crafted-startup (PR #341)
- [x] crafted-ui (PR #336)
- [x] crafted-updates (PR #340)
- [x] crafted-workspaces (PR #338)
- [x] crafted-writing (PR #329)

Note: 
- `crafted-init` and `crafted-early-init` omitted as they are covered by the `getting-started-guide.org` file.
- `crafted-package` omitted as it's covered by the section on alternative package managers.

### RC1
- [x] Transition Guide  (PR #377)
- [x] Modules contributing guide (PR #384)

sthesing:
I'm working on documentation for `rational-completion`. I have two questions regarding which I could use opinions or advice:

1. My current approach is to explain what each of the used packages does by stacking them on top of each other and describe how it changes Emacs' behaviour: how does hitting "M-x" and typing "buf <SPC> swi <TAB>" behave in a fresh install, with vertico, with vertico+orderless, with vertico+orderless+marginalia and so on...
I think that should demonstrate both the specifics of each package as well as how they work together. Can you folks think of any reason not to do it like that?
2. In the org-file, I'd like to include screenshots of the changing behaviour, which obviously won't make it into the info files generated from it. I'll write the text so that it makes sense with or without the screenshots. Is that OK or do the org and info versions of the documentation have to be absolutely identical?

jeffbowman:
My original intention with the org files was to make it easier to get people to contribute as it is more commonly known than is the texinfo markup. I never even considered it might be used to host things like screenshots of examples since the point was to use `org export` to generate the info files.

[...]

PR #155, opened 2022-05-16
Documentation for the module rational-completion

sthesing:
- As mentioned in #128, I try to explain the module by explaining what the used packages do, step-by-step.
- I completely lost an intuitive feeling for the intended audience/level of detail, though. In the end, I try to write the documentation that I'd have wished for as a fairly new Emacs user who is not up to date with the current buzz of the (M)ELPA packaging scene. Any feedback with regard to that choice would be welcome.
- I hope I generated the info-file in the correct way. It looks alright to me, but I don't have much experience with how info-files are supposed to look.
- I also took the liberty to rename the file `mit.org` to `mit-license.org` because its existence confused me until I opened it.


jeffbowman:
Nice... The images even made it into the info file, I didn't think that would happen. Thanks!

PR #167, opened 2022-06-25
Fix doom-modeline initialization

cfclavijo:
`doom-modeline-init` function was removed in commit https://github.com/seagle0128/doom-modeline/commit/11faebc1f9239302eb4c431448aff0143005610b, use `doom-modeline-mode` instead


jeffbowman:
Thanks!

PR #170, opened 2022-07-02
Add rational-osx.el module

saccarosium:
Adds:
* Option to make macOS titlebar transparent;
* Sane configuration for macOS specific modifiers (opt, command);
* Some keybinding involving the command key for making the all experience more
  consistent with macOS default ones

All the new keybindings are bound to hyper, so they don't overwrite the default
ones.


jeffbowman:
Thanks! Just a small request to update the header of the module to match the rest of the modules.

jeffbowman (changes requested):

saccarosium:
Should be right now.

jeffbowman (approved):

PR #176, opened 2022-07-05
Add to rational default some sane behavior of tmp files

saccarosium:
Add supports for the [xdg directories specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), and move all the tmp files into those directories.

### DIFF
--- a/modules/rational-defaults.el
+++ b/modules/rational-defaults.el
@@ -11,6 +11,36 @@
 
 ;;; Code:
 
+;; Defining variable for XDG directories specifications
+(defvar xdg-data-home-directory
+  (if (getenv "XDG_DATA_HOME")
+    (getenv "XDG_DATA_HOME")
+    (expand-file-name ".local/share/")))
+
+(defvar xdg-cache-home-directory
+  (if (getenv "XDG_DATA_HOME")
+    (getenv "XDG_DATA_HOME")
+    (expand-file-name ".cache/")))
+
+;; Keep emacs folder clean
+
+;; Backup file ( init.el~ )
+(customize-set-variable backup-directory-alist `(("." . ,(expand-file-name "emacs/backups/" xdg-data-home-directory))))
+;; Emacs autosave files ( #init.el# )
+;; The autosave feature doesn't create the directory
+(make-directory (expand-file-name "emacs/auto-save/" xdg-cache-home-directory) t)
+(customize-set-variable auto-save-file-name-transforms
+    `((".*" ,(expand-file-name "emacs/auto-save/" xdg-cache-home-directory) t)))
+
+;; eshell history, lastdir and aliases
+(customize-set-variable eshell-directory-name 
+  (expand-file-name "emacs/eshell" xdg-data-home-directory))
+(customize-set-variable eshell-aliases-file 
+  (expand-file-name "emacs/eshell/aliases" xdg-data-home-directory))
+;; bookmark default file
+(customize-set-variable bookmark-default-file 
+  (expand-file-name "emacs/bookmark-default.el" xdg-data-home-directory))
+
 ;; Revert Dired and other buffers
 (customize-set-variable 'global-auto-revert-non-file-buffers t)
 
@@ -18,7 +48,7 @@
 (global-auto-revert-mode 1)
 
 ;; Use spaces instead of tabs
-(setq-default indent-tabs-mode nil)
+(customize-set-variable-default indent-tabs-mode nil)
 
 ;; Use "y" and "n" to confirm/negate prompt instead of "yes" and "no"
 ;; Using `advice' here to make it easy to reverse in custom
@@ -27,7 +57,7 @@
 ;; N.B. Emacs 28 has a variable for using short answers, which should
 ;; be preferred if using that version or higher.
 (if (boundp 'use-short-answers)
-    (setq use-short-answers t)
+    (customize-set-variable use-short-answers t)
   (advice-add 'yes-or-no-p :override #'y-or-n-p))
 
 ;; Turn on recentf mode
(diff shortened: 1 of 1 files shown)


modules/rational-defaults.el
```diff
@@ -11,14 +11,44 @@
 
 ;;; Code:
 
+;; Defining variable for XDG directories specifications
+(defvar xdg-data-home-directory
```
jeffbowman, line 15: we have `rational-config-var-directory` and `rational-config-etc-directory`, so not sure we need additional configuration for yet 2 more folders to spread Emacs files over. If we were not going to use the rational directories, these would certainly make sense, and then you probably would want `(expand-file-name ".local/share/" (getenv "HOME"))` so the full path is set rather than the relative one from wherever Emacs is started.

This may be an opportunity to introduce a `rational-use-xdg-folders` variable and then prefer those over the `rational-config-*-directory` folders.

modules/rational-defaults.el
```diff
@@ -11,14 +11,44 @@
...
+
+(defvar xdg-cache-home-directory
+  (if (getenv "XDG_DATA_HOME")
+    (getenv "XDG_DATA_HOME")
+    (expand-file-name ".cache/")))
+
+;; Keep emacs folder clean
+
+;; Backup file ( init.el~ )
+(customize-set-variable backup-directory-alist `(("." . ,(expand-file-name "emacs/backups/" xdg-data-home-directory))))
```
jeffbowman, line 28: I don't mind the default Emacs behavior for backups and autosaves, so I would personally prefer not to have these. However, I'd probably prefer to use `var/backups` and `var/autosaves` instead (from the `rational-config-var-directory`) of the xdg folders just so everything stays contained in one place.  If there were a `prefer-xdg-paths` var or something as I mentioned in another comment, that could drive this as well.

modules/rational-defaults.el
```diff
@@ -11,14 +11,44 @@
...
+;; Emacs autosave files ( #init.el# )
+;; The autosave feature doesn't create the directory
+(make-directory (expand-file-name "emacs/auto-save/" xdg-cache-home-directory) t)
+(customize-set-variable auto-save-file-name-transforms
```
jeffbowman, line 32: Why change this? The default has a much more robust regex and uses the tmp filesystem which gets cleaned on system reboots and is largely "out of site, out of mind"  already.

modules/rational-defaults.el
```diff
@@ -11,14 +11,44 @@
...
+    `((".*" ,(expand-file-name "emacs/auto-save/" xdg-cache-home-directory) t)))
+
+;; eshell history, lastdir and aliases
+(customize-set-variable eshell-directory-name
```
jeffbowman, line 36: eshell stuff should probably fit in the `rational-config-var-directory` unless we have a `prefer-xdg-paths` type of setting.

modules/rational-defaults.el
```diff
@@ -11,14 +11,44 @@
...
+(customize-set-variable auto-save-file-name-transforms
+    `((".*" ,(expand-file-name "emacs/auto-save/" xdg-cache-home-directory) t)))
+
+;; eshell history, lastdir and aliases
+(customize-set-variable eshell-directory-name 
+  (expand-file-name "emacs/eshell" xdg-data-home-directory))
+(customize-set-variable eshell-aliases-file 
+  (expand-file-name "emacs/eshell/aliases" xdg-data-home-directory))
+;; bookmark default file
+(customize-set-variable bookmark-default-file
```
jeffbowman, line 41: There is currently a draft PR to move other files (bookmarks, saveplace, tramp, etc) to the `rational-config-var-directory`. This probably belongs there as well.

modules/rational-defaults.el
```diff
@@ -11,14 +11,44 @@
...
+
 ;; Revert Dired and other buffers
 (customize-set-variable 'global-auto-revert-non-file-buffers t)
 
 ;; Revert buffers when the underlying file has changed
 (global-auto-revert-mode 1)
 
 ;; Use spaces instead of tabs
-(setq-default indent-tabs-mode nil)
+(customize-set-variable-default indent-tabs-mode nil)
```
jeffbowman, line 51: Probably a typo, no such form `customize-set-variable-default`

modules/rational-defaults.el
```diff
@@ -39,15 +69,15 @@
 (customize-set-variable 'kill-do-not-save-duplicates t)
 
 ;; Make scrolling less stuttered
-(setq auto-window-vscroll nil)
+(customize-set-variable auto-window-vscroll nil)
```
jeffbowman, line 72: `auto-window-vscroll` is not a `defcustom` variable, thus `setq` is probably more appropriate here.

modules/rational-defaults.el
```diff
@@ -39,15 +69,15 @@
...
 (customize-set-variable 'fast-but-imprecise-scrolling t)
 (customize-set-variable 'scroll-conservatively 101)
 (customize-set-variable 'scroll-margin 0)
 (customize-set-variable 'scroll-preserve-screen-position t)
 
 ;; Better support for files with long lines
-(setq-default bidi-paragraph-direction 'left-to-right)
-(setq-default bidi-inhibit-bpa t)
+(customize-set-variable-default bidi-paragraph-direction 'left-to-right)
+(customize-set-variable-default bidi-inhibit-bpa t)
```
jeffbowman, line 80: `bidi-inhibit-bpa` is not a `defcustom` variable, so `setq-default` is more appropriate here.

jeffbowman (changes requested):
While I'm not 100% opposed to XDG paths, the problem is they don't work on Microsoft Windows. So, we might prefer the `rational-config-var-directory` for that reason alone. It might be interesting to consider a `rational-prefer-xdg-paths` type of variable to control using either the `rational-config-var-directory` or the XDG paths for state files. I think people will be a bit lost trying to find their bookmarks in $HOME/.local/share/emacs/bookmarks but maybe not. 

Additionally, there is #158 which is in progress to do some of the same things this one is doing. perhaps we can conflate these two PRs into a single?

saccarosium:
> they don't work on Microsoft Windows

You are right! I always forget about windows support... I get your point.

>  Additionally, there is #158 which is in progress to do some of the same things this one is doing. perhaps we can conflate these two PRs into a single?

Sure, let me close this PR and let me examine #158

PR #179, opened 2022-07-07
Evil docs

vaidyafamily:
Added rational-evil documentation and recreated the info file.


daviwil:
This looks great!  @jeffbowman anything else you think needs to be done here?  If the content isn't 100% accurate yet, we can fix that in a follow-up PR.  I'd like to get all pending PRs merged before the rename in #187.

jeffbowman:
Looks good to me!

PR #182, opened 2022-07-07
Make rational-evil search more like vim

LucasTavaresA:
also removed the unnecessary `rational-evil-vim-muscle-memory` variable as `customize-set-variable` already does this job

### DIFF
--- a/modules/rational-evil.el
+++ b/modules/rational-evil.el
@@ -55,6 +55,9 @@
 (require 'evil)
 (evil-mode 1)
 
+;; Make evil search more like vim
+(evil-select-search-module 'evil-search-module 'evil-search)
+
 ;; Turn on Evil Nerd Commenter
 (evilnc-default-hotkeys)
 


modules/rational-evil.el
```diff
@@ -39,19 +34,20 @@
 ;; Set some variables that must be configured before loading the package
 (customize-set-variable 'evil-want-integration t)
 (customize-set-variable 'evil-want-keybinding nil)
+(customize-set-variable 'evil-want-C-i-jump t)
```
jeffbowman, line 37: The point of the `rational-evil-vim-muscle-memory` variable seemed to be to (by default) leave these at their default state, but if the user wished to turn them on, then these settings would apply. In this code, they are applied by default regardless rather than opted in via the `rational-evil-vim-muscle-memory` value.

Additionally, the `'evil-want-C-i-jump` value is set to `t` here and to `nil` a few lines below.

jeffbowman (changes requested):
The change for removing the rational-evil-vim-muscle-memory should be its own PR, and we should consider if it should be removed.

modules/rational-evil.el
LucasTavaresA: i think that it depends on what rational wants as a whole, its opt in or out?

jeffbowman (approved):

modules/rational-evil.el
saccarosium: I made it opt-in because, even though you are using evil, it doesn't mean that you want evil to overtake some Emacs keybinds. I think it should be a user choice if you want to prioritize vim muscle memory or using vim motion in combination with Emacs ones.

In my opinion the hole point of Rational Emacs is to stick to the defaults, or at least packages that take advantage of built-in behavior, and make some niceties, that go away from the default, opt-in.

I would even say to change `C-d` to be opt-in. But this is not something I can decide alone.

daviwil: Yep, I think opt-in is always a better choice.  Instead of being a "we choose everything for you" distro, this should be more of a "add things you might want quickly" distro.

daviwil: I also agree that `C-u` should be opt-in because it's a critical key binding for Emacs (`universal-argument`) and changing that will break standard Emacs usage patterns.

daviwil:
Thanks @LucasTavaresA!

Issue #153, opened 2022-05-14
rational-compile module always compiles modules, even when no change has been made


jeffbowman:
The `rational-compile` module compiles modules on startup each time Emacs starts. It should only compile modules if the source is newer than the `elc` byte-compiled file. 

cc: @abarocio80

abarocio80:
I am trying to manually verfy if the source file needs compilation, and I learned about `native-compile-async-skip-p`, but it always answers nil to skip compilation.

Is there any issue or bug with this function? Or am I understanding wrong the functionality/interface of this function?

jeastman:
It looks like `native-compile-async-skip-p` check for a few error conditions (`nil` selector, etc.) and if the item is currently being compiled. It does not appear to check if the file is newer than the binary.

One fairly easy solution would be to check the compiled version against the source with `file-newer-than-file-p`.

abarocio80:
Thanks, @jeastman, for the tip.

I already implemeted such function, but it still tries to compile everything on startup, even what I have already compiled.

I belive, the question is "How do we locate the `.eln` compiled file from a source file?" My atempt failed.

This is what I added

```emacs-lisp
(defvar rational-compile--eln-load-path '()
  "Diectories to search in the modules native-compiled files.")
;; [...]
(dolist (path native-comp-eln-load-path)
  (when (and (file-exists-p path)
             (or (string-prefix-p user-emacs-directory path)
                 (string-prefix-p rational-config-path path)))
    (dolist (dir (directory-files path))
      (when (string-prefix-p emacs-version dir)
        (add-to-list 'rational-compile--eln-load-path
                     (expand-file-name dir path))))))
;; [...]
(defun rational-compile-locate-compiled (source)
  "Locates the compiled version of SOURCE in
`rational-compile--eln-load-path'."
  (let ((compiled nil)
        (base (file-name-base source)))
    (if (featurep 'native-compile)
        (dolist (dir rational-compile--eln-load-path)
          (when (file-exists-p dir)
            (dolist (eln (directory-files dir))
              (when (and (not compiled)
                         (string-suffix-p ".eln" eln)
                         (string-prefix-p base eln))
                (setq-local compiled (expand-file-name eln dir))))))
      (let ((elc (concat source "c")))
        (when (file-exists-p elc)
          (setq-local compiled elc))))
    compiled))
;; [...]
(defun rational-compile-file-p (source)
  "Whether to compile SOURCE or not."
[...]

Issue #187, opened 2022-07-09
Rename to Crafted Emacs


daviwil:
I think this is the best-sounding formulation of a name (thanks @jeffbowman!) that fulfills two goals:

- Doesn't sound vaguely elitist
- Fits better with the "brand" of the System Crafters community

To affect the change, I think we should create a new branch where we do all of the module renaming and documentation updates before we flip the switch and rename the entire repo (it should also be `crafted-emacs` once this is done).

I've created a branch called `crafted-rename` to start this work but anyone can send PRs to help!

daviwil:
See also the sibling repo I just created: https://github.com/SystemCrafters/crafted-guix

jeffbowman:
PR #194 submitted for this. Needs review to make sure I didn't goof anything up.

PR #190, opened 2022-07-09
Fix rational-compile for issue #153

abarocio80:
# Fix `rational-compile` for issue #153 
Still have a strange bug with `native-compile`.

If you start your emacs session configured with `rational-compile` to
compile modules, init or config on startup, it will stall on a second
startup, complaining that `rational-package-install-package` is not
available.

The issue seems to be solved once you delete all `.elc` files from
`user-emacs-directory` and `rational-config-path`, and only then
`rational-compile-init`, `rational-compile-config` or
`rational-compile-modules`.

I don't know if that problem is specific to my system/configuration,
but I have tried daviwil's configuration with only `rational-compile`
added to it, and it crashes.

Maybe if `rational-package-install-package` is turned into a module,
and that module is required on each module that uses it, it could stop
complaining that it is not available.

Please read, comment, make suggestions and corrections.


abarocio80:
Just seen issue #187. Will wait if it needs to first go through renaming process.

ajxn:
Is it possible to detect the absence of **rational-package-install-package** and then remove the .elc files as a hack, because it seams like this problem is also in my setup.

I am using latest master.

abarocio80:
My guess is that, as those files got compiled withouth a `rational-package-install-package` function (in the compilation context), don't know what to do when the compiled file is loaded and executed, no matter if the function is already in the execution context.

So, I would propose that each module that make use of `rational-package-install-package` require the package `rational-package` (from #191, now renamed to `crafted-package` and the function `crafted-package-install-package` according to #194).

If some one knows better than this, pelase let me know.

jeffbowman:
@abarocio80 Do you want to take a look into the conflicts with this PR? Maybe once those are resolved we can merge it?

abarocio80:
@jeffbowman, I think I did it.

modules/crafted-compile.el
```diff
@@ -239,7 +245,23 @@ The files to be compiled is defined in
                        (string-equal major-mode "emacs-lisp-mode"))
               (crafted-compile-buffer))))
 
-
+(defun crafted-compile-locate-eln-file (library)
+  ""
```
jeffbowman, line 249: Maybe remove the empty string, or (better) provide some doc here

jeffbowman (changes requested):
Thanks, minor change requested.

abarocio80:
OK. I think it is done.

jeffbowman:
Thanks!

PR #191, opened 2022-07-09
Created a module rational-package to manage emacs' package-managers

abarocio80:
This is an attempt to modularize package installation, according to what I have seen it is been used so far in `rational-emacs`.

It adds support for `straight.el`, and the capacity to add other package-manager systems as backends.

I hope that my desition to put the backends in a subdirectory doesn't offend people. If you have a better solution, I would hear you.

This module tries to aim into the granularity issue mentioned in a live stream.


abarocio80:
Just seen issue #187. Will wait if it needs to first go through renaming process.

abarocio80:
Just realized that I forgot to add the main module to the PR. Sorry.

jeffbowman (approved):
Looks good to me! Thanks!

modules/rational-package.el
```diff
@@ -0,0 +1,44 @@
...
+
+;; Other backends could be added.
+
+;;; Code:
+
+(defvar rational-package-system 'package
+  "What package system to use.
+
+By default, it uses 'package for `package.el'.  Another option is
+'straight for `staright.el'.")
```
daviwil, line 25:
```suggestion
'straight for `straight.el'.")
```

modules/rational-use-package.el
```diff
@@ -17,21 +17,22 @@
...
- (when (version< emacs-version "28.0")
-   (add-to-list 'package-archives (cons "nongnu" "https://elpa.nongnu.org/nongnu/")))
-
- ;; Always add melpa to `package-archives'
- (add-to-list 'package-archives (cons "melpa" "https://melpa.org/packages/"))
-
- (package-initialize)
- (unless package-archive-contents
-   (package-refresh-contents)))
+(when (eq rational-packge-system 'package)
```
daviwil, line 20: We might want to update this part to use the `rational-package-install-package` command since `straight.el`users won't get `use-package` installed correctly without it.

daviwil (approved):
This is a great start, thanks a lot!

jeffbowman (changes requested):

modules/rational-use-package.el
jeffbowman: I think most of this can disappear since we are loading `rational-package` and it configures the stuff here in the case of package.el usage (thus redundant here) or in the case of straight, it isn't needed.

jeffbowman (changes requested):

modules/crafted-package/straight.el
```diff
@@ -0,0 +1,40 @@
...
+
+;; It also adds a couple of macros to make an abstraction of the API:
+;; `rational-package-install-package' and
+;; `rational-package-installed-p'.
+
+;;; Code:
+
+(defvar bootstrap-version)
+(let ((bootstrap-file
+       (expand-file-name "straight/repos/straight.el/bootstrap.el" rational-config-path))
```
jeffbowman, line 21: should be `crafted-config-path` since the rename is now completed

modules/crafted-package/straight.el
```diff
@@ -0,0 +1,40 @@
...
+  (unless (file-exists-p bootstrap-file)
+    (with-current-buffer
+        (url-retrieve-synchronously
+         "https://raw.githubusercontent.com/raxod502/straight.el/develop/install.el"
+         'silent 'inhibit-cookies)
+      (goto-char (point-max))
+      (eval-print-last-sexp)))
+  (load bootstrap-file nil 'nomessage))
+
+(defmacro rational-package-install-package (package)
```
jeffbowman, line 32: s/rational/crafted/ here as well. Also on lines 36, 39, and 40 below.

jeffbowman (changes requested):
minor changes since the rename from rational to crafted.

abarocio80:
Great! Thanks. Let me work on it.

abarocio80:
I think it's done…

jeffbowman (approved):

jeffbowman:
Unfortunately, had to revert this one. 
The problem is the `(require 'crafted-emacs)` in the `early-init.el` file: it doesn't know how to load that file as the `modules` folder is not on the `load-path` yet (it is added in `init.el`). So bootstrapping the package system this way fails and causes Emacs to not load. 

Will investigate options.

jeffbowman:
@abarocio80 please review #201 where I have refactored a bit of this code. Thanks for your effort on this PR!

PR #194, opened 2022-07-10
Crafted rename

jeffbowman:
Renamed rational->crafted everywhere.


daviwil (approved):
Looks good to me!

I wonder if (after merging this PR) we should add some temporary logic to the configuration that will alert the user to the new naming conventions as soon as they open Emacs the next time since they will have to do two things:

- Move their `rational-emacs` config folder to `crafted-emacs`
- Change all `require`s and references in their config files to `crafted`

Bonus points if we can try to do these things for them automatically, but it's possible we could break their config further.

abarocio80:
Will the repo's URL change too? I mean: will it become `https://github.com/SystemCrafters/crafted-emacs`?

jeffbowman:
> Looks good to me!
> 
> I wonder if (after merging this PR) we should add some temporary logic to the configuration that will alert the user to the new naming conventions as soon as they open Emacs the next time since they will have to do two things:
> 
> * Move their `rational-emacs` config folder to `crafted-emacs`
> * Change all `require`s and references in their config files to `crafted`
> 
> Bonus points if we can try to do these things for them automatically, but it's possible we could break their config further.

I added a script to do the conversion. May be able to hook it into `early-init.el` somewhere and run it as an external process if necessary.

rational2crafted.el
```diff
@@ -0,0 +1,120 @@
...
+(defvar crafted-config-path
+  (string-replace "rational" "crafted" rational-config-path)
+  "The user's crafted-emacs config path.")
+
+(print (concat "crafted config path: " crafted-config-path))
+
+;; Copy the rational-emacs folder to crafted-emacs.  This,
+;; effectively, makes a backup of the rational-config-path. From here,
+;; we will only work with in the crafted-config-path folder.
+(copy-directory rational-config-path crafted-config-path)
```
daviwil, line 63: I love that this is a copy instead of a move so we don't totally destroy the original!

rational2crafted.el
```diff
@@ -0,0 +1,120 @@
...
+(defvar crafted-early-config-file
+  (expand-file-name "early-config.el" crafted-config-path)
+  "The user's crafted-emacs early init file, early-config.el")
+
+(print (concat "crafted-config-file: " crafted-config-file))
+(print (concat "crafted-early-config-file: " crafted-early-config-file))
+
+(defun perform-replacement ()
+  (goto-char (point-min))
+  (while (re-search-forward "rational" nil t)
```
daviwil, line 78: I wonder if we should search for `rational-` since all variables, functions, and libraries use a hyphen after?  I know this might not catch other usages of `rational*` but it feels safer to not catch every instance of the word.

README.org
```diff
@@ -10,6 +10,41 @@ pull of the repository. If you are not expecting or prepared to encounter
 such issues, we would recommend you wait for things to stabilize a bit
 before using.
 
+* =TRANSTION FROM AN EXISTING RATIONAL EMACS CONFIGURATION INSTRUCTIONS=
+
+This project was originally called _Rational Emacs_. However, we chose to rename
+to _Crafted Emacs_ to better represent our ideas and fit better with the System
+Crafters goals and messaging.
+
+If you were previously using *rationa-emacs*, please follow these instructions:
```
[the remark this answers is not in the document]
jeffbowman: Probably better that way. I'll make that happen.

PR #208, opened 2022-07-24
initial config for latex and pdf files

thibautbenjamin:
Hi, thanks a lot for your work maintaining this awesome project.
Here is a basic configuration for LaTeX that I use and which works pretty well.

I set it all up using AUCTeX, which albeit not a part of emacs is still a GNU project. I tried to follow the idea of crafted emacs, to stick as much as possible with a minimalistic and non opinionated configuration. I still had some decisions to make along the way, in order to provide a good out-of-the-box experience.

- Because of the quirks of LaTeX, it sometimes need to be compiled several times in a row in order to get everything right. There is a very nice helper script called latexmk, but it is not compatible with AUCTeX out of the box, so I chose to add the package `auctex-latexmk`. Others might not agree with this choice, especially because this package may have a dependency problem on MELPA (it has on one of my system and not in the other), and I made a little workaround that. I do not mind removing this package if people think it's better. 
- I assumed that LaTeX was used to procude pdf files and not postscript. I believe that is what most people use it for anyway, so it is not very opinionated.
- I provided the option to read the pdf inside of emacs, but I added a support for the `pdf-view` mode, except of the native `DocVIew` mode. Again I do not think this is very opinionated, since this package provides much more features such as correlating the source and the output.

On the other hands, here are a few packages and functions that I did not include, as I do not think they are suited to everyone's use of LaTeX, or require a choice that I should not make in place of the user:
- I did not assume that the user has a LSP for latex. There are 2 of them and I do not think I should make the choice instead of the user. If you want me to, I could check if `eglot` is loaded and if one of the language server is installed, and if so add `eglot-ensure` as a hook to `LaTeX-mode`
- I did not use `xenops`, even though I personally love it, because it provides something that I know a lot of people don't care about or don't like
- I did not add any additional package to manage citations. I do not know them well enough. I have had a look at `citar`, mentioned in https://github.com/SystemCrafters/crafted-emacs/issues/130, and I really like it. The fact that it integrates well with `vertico`, `embark`, `marginalia`... makes me think that it could be integrated to `crafted-emacs`, but it is not specific to `LaTeX`. Other kinds of packages exist for citations, and could be added by people who know them better than me.
- I did not set up a spell-checker in LaTeX mode, as I think some people would not like that

### DIFF
--- a/modules/crafted-latex.el
+++ b/modules/crafted-latex.el
@@ -1,0 +1,137 @@
+;;; crafted-latex.el -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2022
+;; SPDX-License-Identifier: MIT
+
+;; Author: System Crafters Community
+
+;; Commentary
+
+;; Sane defaults for LaTeX editing
+
+;;; Code:
+
+(defgroup crafted-latex '()
+  "LaTeX configuration for Crafted Emacs."
+  :tag "Crafted LaTeX"
+  :group 'crafted)
+
+(defvar crafted-latex-latexp
+  "is the latex executable found"
+  :group 'crafted-latex
+  :type 'string
+  :set (executable-find "latex"))
+
+(defvar crafted-latex-latexmkp
+  "is the latexmk executable found"
+  :group 'crafted-latex
+  :type 'string
+  :set (executable-find "latexmk"))
+
+(defcustom crafted-latex-use-pdf-tools
+  "use pdf-tools as the pdf reader
+   (this is automatic if you load `crafted-pdf-reader')"
+  :group 'crafted-latex
+  :type 'boolean
+  :set nil)
+
+(defcustom crafted-latex-inhibit-latexmk
+  "When set to `nil', the package auctex-latexmk gets installed if the
+latex and latexmk executable are found
+
+This package contains a bug which might make it crash during loading
+(with a bug related to tex-buf) on newer systems. For this reason, we inhibit
+the installation of this package by default.
+
+
+If you encounter the bug, you keep this package inhibited. You can install
+a fix (not on melpa) with the following recipe, and the configuration in this file
+will still work
+'(auctex-latexmk :fetcher git :host github :repo \"wang1zhen/auctex-latexmk\")"
+  :group 'crafted-latex
+  :type 'boolean
+  :set t)
+
+
+;; only install and load auctex when the latex executable is found,
+;; otherwise it crashes when loading
+
+(when crafted-latex-latexp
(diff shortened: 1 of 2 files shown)


thibautbenjamin:
Also: I am willing to write down an info module after we agree on the definitive configuration

jeffbowman:
This looks really good, thanks! It's unfortunate the `auclatex-latexmk` maintainer hasn't yet merged the fix to make that project work, I use `latexmk` for all of my projects.  In my case though, I use a `Makefile` and run it from the `project-compile` command.

jeffbowman:
> Also: I am willing to write down an info module after we agree on the definitive configuration

Yes, I think we need this. We may also need to reference the `file-line-error` solution from the `auctex-latexmk` page (under the MikTex section). So, there may need to be a "troubleshooting" section for this particular package to deal with that as well as the `(require 'tex-buf)` problem as well.

thibautbenjamin:
> This looks really good, thanks! It's unfortunate the `auclatex-latexmk` maintainer hasn't yet merged the fix to make that project work, I use `latexmk` for all of my projects. In my case though, I use a `Makefile` and run it from the `project-compile` command.

Yes, the downside with using `latexmk` from a `Makefile` and calling it with `project-compile` is that emacs is not able to pick up on the LaTeX error and give you a correct feedback like AUCTeX does by default. Also I am not sure it is possible to keep the correlation between the source and output, which is something quite important in modern LaTeX editors, especially overleaf.

I will add an info module soon-ish, to document the config a little bit and especially this part with `auctex-latexmk` and the `(require 'tex-buf)` issue. I do not really know how things are supposed to get resolved in this kind of situation, but it seems that the author has gone inactive and I have no idea who should fix the package on MELPA

jeffbowman:
I had to fix a few things, defvar vs defcustom, `crafted-package-install-package` takes a package not a list, minor things, all good in the end. 

Now that I know how to fix the `tex-buf` problem, I'm not too worried about it, I can edit the code after I install it and then hopefully the maintainer will fix the actual package sometime. 

I haven't used the correlation between source and output before, I'm not sure what I'm missing lol.  For errors, I just have become used to trying to figure out the location of errors from the output of the Makefile, so, again, I'm not sure what I'm missing by not using AUCTex before.  I have used overleaf, and the only real problem I have is the amount of time it takes to compile some documents exceeds their limitations for the free account.  I haven't paid for a subscription there because I do almost all of my LaTeX work in Emacs, and only use Overleaf when I need to share it between work/client computers where I don't necessarily have access to Emacs and a LaTeX environment (ala LaTeX Live or similar).

PR #210, opened 2022-07-25
elm module

wereket:
elm programming support


modules/crafted-elm.el
```diff
@@ -0,0 +1,18 @@
...
+;; SPDX-License-Identifier: MIT
+
+;; Author: System Crafters Community
+
+;;; Commentary:
+
+;; elm configuration
+(crafted-package-install-package 'elm-mode)
+
+(setq elm-format-on-save t)
```
jeffbowman, line 13: Prefer `customize-set-variable` here instead of `setq`

jeffbowman (changes requested):

PR #212, opened 2022-07-29
add-to-list doesn't add duplicates so the conditional is unnecessary

zoliky:


jeffbowman:
True, it was mostly there for documentation purposes.

PR #216, opened 2022-08-01
Delete references to removed `modules/crafted-use-package.el`

nspaeth:
`modules/crafted-use-package.el` was deleted in commit #b331d44 but was still referenced in the readme and examples.


jeffbowman:
Thanks!

PR #221, opened 2022-08-19
Updates a few references from Rational to Crafted

haroon-sheikh:
Just updating a few references from Rational to Crafted.


jeffbowman:
Thanks for finding those!

PR #223, opened 2022-08-20
Fix early-init typo "crafted-boostrap-directory"

christoffer-arvidsson:
Found a few typos when looking through early-init.


jeffbowman:
Whoops!!! Thanks for finding those.

PR #229, opened 2022-09-11
fix typo in cljr-add-keybindings-with-prefix

petervlk:
This simple PR fixes a typo which prevents intended rebind from being applied.


jeffbowman:
Well, shucks... all for the want of an 'r'!  Thanks for catching that!

PR #238, opened 2022-10-05
[compile] fix issue when the cache is not found

thibautbenjamin:
The module crafted-compile was causing an error on startup, because the cache for the file `config.el` was not found. I changed it so it will always perform the compilation in that case


modules/crafted-compile.el
```diff
@@ -116,13 +116,12 @@ directory will be compiled, but not it's subdirectories."
...
-    (dolist (source f)
-      (if (file-newer-than-file-p
-             (crafted-compile-locate-eln-file (file-name-base source))
-             source)
-          (native-compile-async f)
-        (message "Skipping compilation of file %s" source)))
-
+      (dolist (source f)
+        (let ((cache (crafted-compile-locate-eln-file (file-name-base source))))
+          (if (or (eq cache nil)
```
jeffbowman, line 121: prefer to use `null` here, ie `(if (or (null cache)...`

jeffbowman (changes requested):
minor change

PR #239, opened 2022-10-06
fix crafted-ensure-package on guix systems

emturner:
# Background
I was trying to use `crafted-ensure-package` to make sure I'd correctly installed `paredit` through *guix home*, but despite installing it correctly, (I could see `emacs-paredit` in my installed packages list in my current home generation), I was still getting a message on startup saying it was missing.

`paredit` was available, however, under `M-x`.

# Investigation
I think the problem is the following: when a package is installed through guix, it appears the feature is made available under '<package-name>-autoloads', rather than '<package-name>'.

This then causes the `(featurep package)` check to fail in `crafted-ensure-package`, when `crafted-prefer-guix-packages` is enabled.

### DIFF
--- a/init.el
+++ b/init.el
@@ -44,14 +44,18 @@
 (customize-set-variable 'visible-bell 1)  ; turn off beeps, make them flash!
 (customize-set-variable 'large-file-warning-threshold 100000000) ;; change to ~100 MB
 
+(defun guix-package-to-feature-name (package &optional args)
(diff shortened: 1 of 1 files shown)


emturner:
Playing around with it further today, appears that some packages end up with the `-autoloads` extension, but other's (e.g. `tuareg`) don't.

Unsure if there's an easy way to work out which one it might be

PR #241, opened 2022-10-10
add falback completion style

petervlk:
This PR adds a fallback completion style as recommended by author of orderless in the [projects docs](https://github.com/oantolin/orderless/blob/6b86527b30ef96e047d97e314ac640a410891d1f/orderless.texi#L122).

It helps to solve [completion issue](https://github.com/clojure-emacs/cider/issues/3019) where namespace prefix is being ignored.

The proposed solution was found [here](https://github.com/clojure-emacs/cider/pull/3226).


jeffbowman:
Nice change, thanks for the PR!

PR #245, opened 2022-10-29
Add crafted-workspaces module

### LINKED ISSUE #233: Use tabspaces to implement crafted-workspaces
Alongside consult with 'b' and 'p' narrowing keys this is kind of a one-liner.

In my config I default to only listing buffers in tabspace, then use the 'b' narrowing key in consult to see all buffers. 

I suppose crafted-workspace would also want default keybindings too. That would make this more than a one-liner.

Cerebus:
- Adds tabspaces.el with simple config and one workaround
- Closes #233


jeffbowman:
Thanks!

PR #249, opened 2022-11-22
Invoke eldoc after corfu selection

Cerebus:
Snagged this hint from [here](https://github.com/minad/corfu/issues/80#issuecomment-970588008), seemed globally applicable since `global-eldoc-mode` is `t`.


jeffbowman:
Nice! Thanks!!

PR #256, opened 2022-12-03
Add corfu-terminal to when non-graphical display in use.

Cerebus:
[Fixes an early complaint.](https://github.com/SystemCrafters/crafted-emacs/pull/86#issue-1148509706)


jeffbowman:
Thanks! I had forgotten about that complaint.

PR #257, opened 2022-12-03
Add documentation on working around eglot with flymake.

### LINKED ISSUE #250: Eglot stomps on flymake-diagnostic-functions
When eglot takes over a buffer, it replaces all entries in flymake-diagnostic-functions.  If your LSP has linting, you won't notice, but with LSPs like pyright you lose the mode's built in diagnostics.  

For python-mode, I'm currently using the following, and it seems to work.

```emacs-lisp
(defun my/eglot-python-fixup ()
  (add-hook 'eglot-managed-mode-hook (lambda () (add-hook 'flymake-diagnostic-functions #'python-flymake t t)))
  )

(add-hook 'python-mode-hook #'my/eglot-python-fixup)
```

Adapted from [this](https://www.reddit.com/r/emacs/comments/s96wec/comment/ii6norq/?utm_source=share&utm_medium=web2x&context=3).

A general fix that captures the mode's value and adds it to back to flymake-diagnostic-functions after the call to eglot-ensure is probably possible, but I haven't tried this yet.

Cerebus:
Closes #250.

### DIFF
--- a/docs/crafted-ide.org
+++ b/docs/crafted-ide.org
@@ -14,3 +14,25 @@
    #+begin_src emacs-lisp
      (customize-set-variable 'eglot-autoshutdown nil)
    #+end_src
+
+** Workarounds
+
+   When eglot initializes management of a buffer, it overwrites
+   =flymake-diagnostic-functions=.  However, Not all LSP servers support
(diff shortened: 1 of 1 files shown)


jeffbowman:
Thanks for the PR!

PR #279, opened 2023-06-06
Craftedv2beta

japhir:
After trying it out with a pretty bare install from scratch (time to clean up my own config a lot) I rand into some consistency issues with the version checks for emacs in different modules. I've changed them throughout to rely on the variable `emacs-major-version`, which has been introduced at or before emacs 19.23 (so should be safe to use?).

I hope this is correct :).

### DIFF
--- a/bootstrap/crafted-package-bootstrap.el
+++ b/bootstrap/crafted-package-bootstrap.el
@@ -28,7 +28,7 @@
 ;; Emacs 27.x has gnu elpa as the default
 ;; Emacs 28.x adds the nongnu elpa to the list by default, so only
 ;; need to add nongnu when this isn't Emacs 28+
-(when (version< emacs-version "28")
+(when (< emacs-major-version 28)
   (add-to-list 'package-archives '("nongnu" . "https://elpa.nongnu.org/nongnu/")))
 (add-to-list 'package-archives '("stable" . "https://stable.melpa.org/packages/"))
 (add-to-list 'package-archives '("melpa" . "https://melpa.org/packages/"))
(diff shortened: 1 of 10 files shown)


japhir:
It's not the same issue so I should have probably made a separate branch, but this now also takes care of #280

jeffbowman:
> It's not the same issue so I should have probably made a separate branch, but this now also takes care of #280

yep, a separate branch would be appreciated to keep things cleaner.

jeffbowman:
Thanks for the PR!  Some thoughts:

I was pretty consistent using `version<` for almost everything.  There were a few cases where I wanted to check that the version of Emacs was "version 29 or later", which is where you saw the calls to `string-to-number`, when I should have written `(version< "29" emacs-version)` which returns `t` when Emacs version is "29.1" (or even "29.0.50" or "29.0.91", etc).  That would have been consistent as well.  The reason for the strings, `version<` requires them and it is the function designed to handle the `emacs-version` variable - which is why I chose to use it.  The places where things go sideways with your implementation are lines like this one: `(when (> emacs-major-version 29)` which will fail when now (with Emacs 28.x) and again when Emacs 29 is released (ie 29 is not greater than 29).  There are a few other conditions where the approach of only checking the `emacs-major-version` numerically will not work out doing what we need it to.   Therefore I'm going to close this PR for that reason and I'll go back and fix the places where I should be using `version<` - so huge thanks for finding those and taking the time to give this some needed attention!

The second reason I am closing this PR without merging it is for the `lambda` fixes you provide.  If you would submit that again as it's own PR, that would be great!! 

Thank you very much!!

Issue #284, opened 2023-06-19
[craftedv2beta] crafted-emacs.org does not contain a Modules section


jvdydev:
Figured I'll open a new issue for that as I'd tackle `crafted-org` documentation first (see #283 for context).

The `CONTRIBUTING.org` file on how to add modules:
```
Add an Org file named for the module. 
Once you are ready to include it in the =crafted-emacs.org= file, add it to the modules section [...]
```

It seems that during the cleanup of the beta branch, the `Modules` section got removed from `crafted-emacs.org`.
I can re-add it before working on the crafted-org docs, just figured I'll open it separately as it affects everyone else looking into contributing module documentation.

Issue #283, opened 2023-06-14
[craftedv2beta] crafted-org contains both denote and org-roam


jvdydev:
More of a discussion question/point I noticed while trying out the v2 beta:

The new `crafted-org` module contains package definitions for both `org-roam` as well as `denote`.
While not identical, they provide similar-enough functionality (as pointed out in the comments in `crafted-org-packages.el`) that most people will probably not use both.

`org-roam` requires a C compiler to compile the SQLite binary (and by extension another thing running).
For the Linux/BSD users that's usually not a problem, but other OS'es or situations may not allow (or require additional setup for) the installation of a C compiler (not sure, but probably affects for example #274).
Not sure how much the SQLite usage as a whole eats into the idea of keeping everything "close to built-in".

`denote` on the other hand offers a lot less comfort functions (e.g. no built-in equivalent to `org-roam-node-find`).

My proposal would probably be to have denote as a default and add documentation on org-roam (including the additional requirement(s) and maybe a simple example configuration).

jeffbowman:
This is a good suggestion.  `denote` lines up more closely with the principles of this project.   `org-roam`, on the other hand, is widely used, so I thought putting both in the configuration would satisfy more users.  

In Emacs 29, there is an option to build Emacs with `sqlite`, so the C compiler isn't needed later, only the libraries which may need to be installed by the user independently of Emacs.  At least that is how I understand it.  That doesn't solve the problem for Android users per se, but for users on other platforms, they shouldn't need a C compiler (when using the (hopefully) imminently release Emacs 29).  For users of Emacs prior to version 29, though... well, that's a different story. 

I wonder if you'd be willing to provide a PR to support your suggested change?

Thanks for the suggestion!

jvdydev:
I'd be happy to work on this, however I think providing docs for crafted-org would also make sense (while also removing `org-roam` and it's related configuration from the actual module).
Should I work on that as a whole or split it out into two PRs?

jeffbowman:
Thank you very much!  I look forward to your work! I would prefer separate PRs if that isn't too much of a hassle on you.

jvdydev:
I experimented around a bit more in a container regarding the C compiler dependency and it seems that even with the current emacs-29 branch, a C compiler is still required for org-roam.

That may change when Emacs 29 is more readily available and org-roam gets an update to use the internal sqlite interface.

In any case, changes are in the work, I'll make the PRs soon.

jeffbowman:
[...]

PR #285, opened 2023-06-23
[craftedv2beta] Module documentation: crafted-org

jvdydev:
Addressing #284 and the documentation side of #283.

For the modules section, I re-phrased the one from current master for v2.
The crafted-org documentation is already assuming the removal of `org-roam` as per #283.


jeffbowman (changes requested):
Minor change to correct the configuration `require` form.

jeffbowman (approved):

PR #286, opened 2023-06-23
Remove org-roam from crafted-org module

jvdydev:
As discussed in #283.
Configuration is already moved over into documentation (see #285).


jeffbowman:
Looks good, thanks!

PR #287, opened 2023-06-24
Fix broken link in Getting Started Guide (Early Init File)

jvdydev:
The link to the early init file info page was broken. Fixed and regenerated info file.


daviwil:
Awesome, thanks!

PR #289, opened 2023-06-24
Fix unreachable require statements

sthesing:
## Observation

In a fresh install, when a user has
```
(require 'crafted-completion-config)
```
in their init.el, many parts of the config silently fail to load.

## Reason

The package specific configs are behind `(when (featurep` conditions. These evaluate to nil because `package-install-selected-packages` does _install_ the packages, but doesn't _require_ them.

The necessary require statements to make these conditions true is _inside_ the statement, e.g.:
```
(when (featurep 'vertico)
  (require 'vertico)
  ...)
```
So unless the user doesn't require the respective package somewhere else, the config silently fails to to take effect.

## Proposed fix

I propose to move `require`s for all packages contained in `crafted-completion-packages` behind a `(when (featurep 'crafted-completion-packages` condition. That way, the combination of `'crafted-completion-packages` and `crafted-completion-config` works as expected.
Also, users who don't want to use the whole `crafted-completion-packages` module can still make use of the `crafted-completion-config` module or one of the contained packages, if they install and require it manually.

## Other module pairs
The phenomenon described here is true for all module pairs `crafted-*-packages` & `crafted-*-config`.
The PR covers only `crafted-completion-packages` & `crafted-completion-config`.
If you like my proposed solution, I'm happy to prepare a PR for the other module pairs.

### DIFF
--- a/modules/crafted-completion-config.el
+++ b/modules/crafted-completion-config.el
@@ -1,4 +1,4 @@
-;;; crafted-completion.el --- Crafted Completion Configuration -*- lexical-binding: t; -*-
+;;; crafted-completion-config.el --- Crafted Completion Configuration -*- lexical-binding: t; -*-
 
 ;; Copyright (C) 2022
 ;; SPDX-License-Identifier: MIT
@@ -13,7 +13,19 @@
 ;; narrowed list of possible choices.
 
 ;;; Code:
-
+(when (featurep 'crafted-completion-packages)
+  (require 'vertico)
+  (require 'vertico-directory)
+  (require 'marginalia)
+  (require 'consult)  
+  (require 'orderless)
+  (require 'embark)
+  (require 'embark-consult)
+  (require 'corfu)
+  (unless (display-graphic-p)
+    (require 'corfu-terminal))
+  (require 'corfu-popupinfo)
+  (require 'cape))
 
 (defun crafted-completion/minibuffer-backward-kill (arg)
   "Delete word or delete up to parent folder when completion is a file.
@@ -30,8 +42,6 @@ ARG is the thing being completed in the minibuffer."
 
 ;;; Vertico
 (when (featurep 'vertico)
-  (require 'vertico)
-  (require 'vertico-directory)
   ;; Cycle back to top/bottom result when the edge is reached
   (customize-set-variable 'vertico-cycle t)
 
@@ -51,7 +61,6 @@ ARG is the thing being completed in the minibuffer."
 
 (when (featurep 'marginalia)
   ;; Configure Marginalia
-  (require 'marginalia)
   (customize-set-variable 'marginalia-annotators
                           '(marginalia-annotators-heavy
                             marginalia-annotators-light
@@ -69,7 +78,6 @@ ARG is the thing being completed in the minibuffer."
 ;;; Orderless
 (when (featurep 'orderless)
   ;; Set up Orderless for better fuzzy matching
-  (require 'orderless)
   (customize-set-variable 'completion-styles '(orderless basic))
   (customize-set-variable 'completion-category-overrides
                           '((file (styles . (partial-completion))))))
@@ -77,7 +85,6 @@ ARG is the thing being completed in the minibuffer."
 
 ;;; Embark
 (when (featurep 'embark)
-  (require 'embark)
 
   (keymap-global-set "<remap> <describe-bindings>" #'embark-bindings)
   (keymap-global-set "C-." 'embark-act)
@@ -86,19 +93,14 @@ ARG is the thing being completed in the minibuffer."
   (setq prefix-help-command #'embark-prefix-help-command)
 
   (when (featurep 'embark-consult)
-    (require 'embark-consult)
     (with-eval-after-load 'embark-consult
       (add-hook 'embark-collect-mode-hook #'consult-preview-at-point-mode))))
 
 
 ;;; Corfu
 (when (featurep 'corfu)
-  (require 'corfu)
-
-  (unless (display-graphic-p)
-    (when (featurep 'corfu-terminal)
-      (require 'corfu-terminal)
-      (corfu-terminal-mode +1)))
+  (when (featurep 'corfu-terminal)
+    (corfu-terminal-mode +1))
 
   ;; Setup corfu for popup like completion
   (customize-set-variable 'corfu-cycle t)                 ; Allows cycling through candidates
@@ -109,7 +111,6 @@ ARG is the thing being completed in the minibuffer."
 
   (global-corfu-mode 1)
   (when (featurep 'corfu-popupinfo)
-    (require 'corfu-popupinfo)
 
     (corfu-popupinfo-mode 1)
     (eldoc-add-command #'corfu-insert)
@@ -119,10 +120,8 @@ ARG is the thing being completed in the minibuffer."
 
 
 ;;; Cape
-
 (when (featurep 'cape)
   ;; Setup Cape for better completion-at-point support and more
-  (require 'cape)
 
   ;; Add useful defaults completion sources from cape
   (add-to-list 'completion-at-point-functions #'cape-file)


sthesing:
An alternative solution would be to have a function that uses the list `package-selected-packages` and requires all of them, right after calling `package-install-selected-packages`.

jvdydev:
I think the "issue" is by design here.

From the getting started guide:
```emacs-lisp
(require 'crafted-completion-packages)  ; must come first to install
                                        ; the packages

(package-install-selected-packages :noconfirm) ; install the packages
                                               ; from the bundle
                                               ; mentioned above.

(require 'crafted-completion-config)
```
I argue this should be made clear that loading any `-packages` modules after the call to `package-install-selcted-packages` will not install the packages, therefor not install and/or configure them and loading a `-config` before installing the package will not load the configuration etc.
The wording of `-packages` requires "installing" packages might be wrong here, instead going for "queuing for installation" might make more sense.

Regarding this:
```markdown
Also, users who don't want to use the whole crafted-completion-packages module can still make use of the crafted-completion-config module or one of the contained packages, 
if they install and require it manually.
```

That is already possible by adding only the wanted packages into their `init.el` instead of requiring the `-packages`, then proceeding as normal, for example to only use vertico:
```emacs-lisp
(add-to-list 'package-selected-packages 'vertico)
(require 'crafted-completion-config)
```
This should (from my understanding of the v2 design) work for every package (and otherwise be considered its own bug).

Comparing that to your commit in this PR would actually partially break modules.
It works for e.g. vertico as `vertico-mode` is an autoload that will require the package for you, but may break packages where the config uses values etc. that are not an autoload.

I think there is potential to expand the `Getting Started Guide` with pointing users to what the built-in functionality that we're using does here.

Also, from the `featurep` help buffer:
```
Use this to conditionalize execution of lisp code based on the
presence or absence of Emacs or environment extensions.
Use ‘provide’ to declare that a feature is available.
```
Meaning `featurep` looks if an extension is available (read: *installed*, found in the load path, not necessarily loaded yet).
`require` then loads the feature.
So I don't quite understand your comment on that part, it works quite well if the package is installed (see ordering above and in the getting started guide).

jvdydev:
I may have to redact my statement on the featurep, I think this can be reduced to less packages however.
The point still stands that a feature is declared by `provide`, then that `provide` is testable via `featurep` and the feature is loaded with `require`.

Why the consult etc. packages don't behave appropriately should probably be investigated separately.

Another issue I find with your proposed iterative fix (the `dolist`) is that not every package needs a require (e.g. keycast-mode in the `screencast` module never gets required).

But all of what I say means little, we'll see what Jeff or David have to say :smile:

jvdydev:
Note: 
I did some earlier comments which prove that pre-coffee testing does not work.
I removed them to avoid confusion.

From testing it seems that you partially are correct that,
at least for the completion package, some configuration fails to load 
(notably binding `consult-line`, as you iterated in IRC/Matrix).

I'm not sure if your proposal covers all the basis however.
If I (a user), decide to *only* install one package of the set 
and not require `crafted-completion-package`, this *may* fail.
It's not guaranteed it fails, mainly because not all packages require to be required.

Your solution actually didn't run the consult configuration for me,
when I only specifically installed consult using:
```
(add-to-list 'package-selected-packages 'consult)

(require 'crafted-completion-config)
```
Which makes sense since `crafted-completion-packages` is not a feature provided in this case.

I think the problem is more nuanced.
`featurep` is **supposed** to show if a feature is loadable, 
at least according to Emacs' Help buffer on featurep (or my understanding of it):
```
Use this to conditionalize execution of lisp code based on the
presence or absence of Emacs or environment extensions.
Use ‘provide’ to declare that a feature is available.
```
Meaning if a package properly *provides* a feature, it is *available* (featurep -> t) and it can be loaded with `require`.

However, that seems to not be the case in quite a few (all?) instances, e.g. (from my testing):
- consult, which always returns `featurep` nil until required
- vertico, which instead is loaded by running the autoload `vertico-mode`
- probably some other scenarios that I haven't looked into yet

I think the better methodology in general would be to see if the package is installed, 
e.g. with `package-installed-p` which is also more clear of what is actually happening.

But all of what I say means little, we'll see what Jeff or David have to say :smile:

sthesing:
> Note: I did some earlier comments which prove that pre-coffee testing does not work. I removed them to avoid confusion.

Thanks for testing it, though. And to be fair, my PR proves that late-night writeing of PRs isn't a good idea, either. Much of the initial confusion is probably due to me forgetting to mention that - of course - I ran `package-install-selected-packages` between loading the two modules in question. My bad, sorry.

> Your solution actually didn't run the consult configuration for me, when I only specifically installed consult using:
> 
> ```
> (add-to-list 'package-selected-packages 'consult)
> 
> (package-install-selected-packages :noconfirm)
> 
> (require 'crafted-completion-config)
> ```

That's right, it only works if the user installs and requires it manually.

> I think the better methodology in general would be to see if the package is installed, e.g. with `package-installed-p` which is also more clear of what is actually happening.

Yes, that is indeed the better solution! I wasn't even aware of `package-installed-p`. Lesson learned, thanks to you!

> But all of what I say means little, we'll see what Jeff or David have to say smile

So, if @jeffbowman and @daviwil agree that this is the way to go, I'll gladly amend the PR.

jvdydev:
No worries.

I think `package-installed-p` limits the compatibility with `package.el` replacement solutions (e.g. `straight.el`).
Not really sure what the best solution here is, I could see two possibilities:
- hope the user installed the package and check if it's in the selection, like
  ```emacs-lisp
  (when (member 'vertico 'package-selected-packages)
  ;; run configuration
  )
  ```
- make a custom variable like `crafted-package-installed-predicate` (from `crafted-package-config.el`) to use (or even re-use that one)

Just some additional thoughts I had while making food :smile:

jeffbowman:
Crafted Emacs v2 prefers `package.el` as the package manager.  So we don't consider how other package managers may or may not load packages.  This keeps things "simple" in the sense that we aren't trying to jump through hoops to setup macros or variables for any package manager that may or may not exist now or in the future.  That said, we do prefer to keep things as generic as we can so other package managers can be used if the user prefers. 

I say all that to set the context for the rest of this discussion; which is simply that we will take a bit of a deeper dive into how `package.el` works in this case and what might be a work-around for the situation here.  We did run into an issue on stream which we solved generically - more on that in a moment. 

When Emacs is starting, it eventually needs to load a package manager.  This should be obvious, and in the absence of other configuration, will load `package.el` and call `package-initialize` to "load" (not the same as the `load` function) the installed packages.  This will setup the `package-alist` with descriptors of all the packages installed (there is a caveat to that statement which we can safely ignore as most likely 99% of the people don't actually know about it).  This process will eventually call `package--load-files-for-activation` which does what it sounds like: it loads the functions marked with `###autoload` and makes sure the `load-path` is setup correctly.   This is **not** the same things as calling `require`.  Think of this more as automatically evaluating the `###autoload` forms in a way that their execution causes the rest of the file to be loaded on execution (a bit of a technical inaccuracy here, but we'd have to go even deeper into the guts of Emacs and how it handles loading things lazily which will take us into the C code and the lisp machine itself... we'll just gloss over this for the moment).  So the difficulty is that a `feature` may not be `featurep` because (at least) one of the autoload functions for that `feature` have not been called, so the `provide` line in that file has not executed and the `feature` is not added to the list `features`, which means the following check fails:

```emacs-lisp
(when (featurep 'vertico)
   ;; ...
)
```

could fail.  We solved this on stream by checking for an autoload function, in this case we updated the code like this:

```emacs-lisp
(when (or (featurep 'vertico) (fboundp 'vertico-mode))
  ;; ...
)
```

This solution worked when David created a configuration using guix-home to install Emacs packages and then configure them simply by using `(require crafted-completion-config)`.  This approach has the benefit of being generic with the cost of having to know something about the package we are checking for - ie we have to know the name of an autoloaded function to check for.

So, why not just use `require`?  The answer is startup performance impact. For some people, that is irrelevant, they start Emacs as a daemon (maybe) and just leave it running... forever.  So, they are willing to suffer the cost of a 30 second (or longer) startup time since it only happens once.  Other people do care how quickly Emacs starts, possibly because they start and stop Emacs several times.  I prefer to approach the problem by optimizing toward quicker start times to try to satisfy both camps.  

In the context of this PR, the code was intentionally written to allow for a granular configuration.  Thus if the user does not want to configure embark they may choose to copy the `crafted-completion-packages` and remove embark from the list then just use `crafted-completion-config` to configure whatever remained.  If embark were not installed, the `(require 'embark)` would fail since the package is not installed (for this reason, I cannot merge this PR).  We could check for `package-installed-p`, but that would fail when the user chose to use some other approach to package management.  I've been looking around to see if there is an elegant and package manager agnostic way to approach this.  The simplest seems to be the `(or (featurep ...) (fboundp ...))` approach, and kinda begs for a macro to keep the solution in one place.  This would be relatively simple to write as a macro, but now we have to know at least one autoload function _a priori_ which adds cognitive burden to the modules.  

I'm still considering approaches, so if either of you (or anyone else lurking on this PR) have ideas, lets talk about them.  I won't migrate Crafted Emacs v2 beta to master without a solution in place.  So, I appreciate this PR and the discussion and testing @sthesing and @jvdydev  have been providing.

jeffbowman:
> ... elided ...
> * make a custom variable like `crafted-package-installed-predicate` (from `crafted-package-config.el`) to use (or even re-use that one)
> 

This is challenging because the `crafted-package-config` module is not pulled in via a `require` anywhere; it, like the other modules, is an optional tool provided to the user.

Issue #288, opened 2023-06-24
[craftedv2beta] Use org-indent-mode in crafted-org


jvdydev:
Noticed that `org-indent-mode` is not activated by default (part of org).

I personally think it helps a lot with making org more naturally readable,
without having to micro-manage indentation oneself.

Activation could be in `crafted-org-config.el`:
```emacs-lisp
(add-hook 'org-mode-hook #'org-indent-mode)
```

jeffbowman:
Not opposed.  PR?

PR #290, opened 2023-06-26
[craftedv2beta] Adding org-indent-mode to crafted-org

jvdydev:
Addressing #288.
Also updated the docs and regenerated the info file.


jeffbowman:
Looks good, thanks for also keeping the docs up-to-date!

PR #292, opened 2023-07-02
add some customization options to splash screen

bacchist:
Minor addition to the splash screen letting users change the theme and font right away.  Mostly for convenience, but changing the font is important for accessibility for users with high DPI screens.


jeffbowman:
Thanks!

PR #294, opened 2023-07-04
Turn off interfering modes from crafted-defaults-config

sthesing:
Users who use both `crafted-defaults-config` and `crafted-completion-config` find unexpected behaviour: both the built-in fido-vertical-mode and vertico try to populate the minibuffer. Same with icomplete and orderless.

I think it's safe to assume that users who load `crafted-completion-config` do not want to use the built-in modes, so they can be turned off here.


jeffbowman:
> I think it's safe to assume that users who load crafted-completion-config do not want to use the built-in modes, so they can be turned off here.

It is not safe to assume this.  If the user did indeed install `vertico` they probably mean to use that package rather than the built-in `fido` or `icomplete` libraries.  If the user instead pulled the package list from the `crafted-completion-packages` module then elided the `vertico` package they might end up with a list like this in their `init.el`:

    (add-to-list 'package-selected-packages 'cape)
    (add-to-list 'package-selected-packages 'consult)
    (add-to-list 'package-selected-packages 'corfu)
    (add-to-list 'package-selected-packages 'corfu-terminal)
    (add-to-list 'package-selected-packages 'embark)
    (add-to-list 'package-selected-packages 'embark-consult)
    (add-to-list 'package-selected-packages 'marginalia)
    (add-to-list 'package-selected-packages 'orderless)

Thus, if they did include the `crafted-defaults-config` they might prefer the `fido` or `icomplete` libraries.  It is also possible they did something like this:

    ;; assume the package list from above was used here as well
    ;; but in this case, the user did not use the crafted-defaults-config
    (fido-vertical-mode 1)
    (require 'crafted-completion-config)

Which, with this patch, would suddenly turn off their completion preference. 

However, you point is well made, we should consider turning off things that  might conflict.  I made a couple suggestions in the PR.  One more thing to consider:

    (require 'crafted-completion-config)
    (require 'crafted-defaults-config)

How do you propose to resolve this situation, which would initially turn off `fido` and `icomplete` libraries, on line 3, but then turn them back on on line 4.  A couple of ideas:

1. Update documentation and suggest using the `crafted-defaults-config` earlier than all other Crafted modules.
2. Check to see if `crafted-completion-config` is being used and if the `vertico` package is being used before turning on either `fido` or `icomplete`. 
3. Maybe both 1 & 2
4. Don't turn on `fido` or `icomplete` in the `crafted-defaults-config` module and allow the user to turn it on manually in their own config.

jvdydev:
Just throwing in my two cents into the conversation:

I don't think imposing a load-order is necessarily useful (and adds indirect coupling between the modules, while the goal is to have fully independent modules).

I could see a two-part solution (assuming we do want to activate `fido/icomplete` for the user):
```emacs-lisp
;; crafted-emacs-completion-config.el
(when (featurep 'vertico) ; this needs updating anyway
  ;; ...
  ;; turn off fido/icomplete
  )

;; crafted-emacs-defaults-config.el
(unless (featurep 'vertico)
  ;; turn on fido/icomplete
  )
```

This way:
- `fido/icomplete` do not activate after `vertico`
- if `vertico` is loaded after, `fido/icomplete` get disabled again

Being linked to `vertico` configuration directly would mean that it doesn't matter from where `vertico` is pulled in either (although the configuration guard needs to be changed anyway ~> #289).

jeffbowman:
Thanks @jvdydev, good ideas.  

Yes the guards around (at least) vertico need to be reworked.  Based on a suggestion after discussion in #289 `locate-library` might be an option to resolve if the package has been installed, which would work for the `crafted-completions-config` module, while `featurep` would work for the `crafted-defaults-module`, although there may still be issues in correct detection.

sthesing:
> It is not safe to assume this. 

True. As always, I'm looking at this from a perspective of relatively new Emacs users (or those who come from other "distributions"). I believe a user who just picks the respective modules "as a whole" would expect them to work out of the box without having to manually turn certain things off. While a user who cherry-picks certain things out of the packages could be considered a "tinkerer", "crafter" or an "advanced user" and they probably wouldn't be surprised if they have to  tweak things a bit.

> How do you propose to resolve this situation, which would initially turn off `fido` and `icomplete` libraries, on line 3, but then turn them back on on line 4. A couple of ideas:
> 
>     1. Update documentation and suggest using the `crafted-defaults-config` earlier than all other Crafted modules.
> 
>     2. Check to see if `crafted-completion-config` is being used and if the `vertico` package is being used before turning on either `fido` or `icomplete`.
> 
>     3. Maybe both 1 & 2
> 
>     4. Don't turn on `fido` or `icomplete` in the `crafted-defaults-config` module and allow the user to turn it on manually in their own config.

Out of these, 3 is the way to go, I think. But I do agree that @jvdydev 's idea is probably better.

> Yes the guards around (at least) vertico need to be reworked. Based on a suggestion after discussion in #289 `locate-library` might be an option to resolve if the package has been installed, which would work for the `crafted-completions-config` module, while `featurep` would work for the `crafted-defaults-module`, although there may still be issues in correct detection.

That suggestion on Mastodon incidentally was also made by me, so if you also think this isn't slowing down startup time too much, we could go for:

```elisp
;; crafted-emacs-completion-config.el
(when (locate-library "vertico")
  ;; ...
  ;; turn on vertico
  ;; turn off fido/icomplete
  )

;; crafted-emacs-defaults-config.el
(unless (and (featurep 'crafted-emacs-completion-config) (locate-library "vertico"))
  ;; turn on fido/icomplete
  )
```
That way it would work as expected for users who just load one or both of these modules and hope they'll work. 
For the rest cherry pick, the chance to meet their expectations should at least be reasonably high ;-)

jeffbowman:
> That suggestion on Mastodon incidentally was also made by me

I thought that might be the case, but I didn't want to assume.  Thanks for the suggestion!!  Also, it looks fine to use `locate-library` to detect if a package is installed.  In theory it is slower since it has to look through the `load-path`, however, I couldn't prove it was indeed slower.  So, this looks like it might be a good way forward.  @sthesing if you are a guix user, it might be useful to test the scenario of installing the Emacs packages with `guix` (perhaps ala `guix-home` or something similar) and then we can test which packages need the `locate-library` guard.  I'd rather not blindly apply that guard just on general principles.

sthesing:
> @sthesing if you are a guix user, it might be useful to test the scenario of installing the Emacs packages with `guix` (perhaps ala `guix-home` or something similar) 

I'm not a Guix user, but I did a quick install in a VM. And yes, locate-library correctly detects Emacs packages installed via Guix.

> and then we can test which packages need the `locate-library` guard. I'd rather not blindly apply that guard just on general principles.

Do you mean that we figure out which packages need a `locate-library` guard as opposed to a `featurep` guard? Or which packages need a guard at all?

jeffbowman:
> Do you mean that we figure out which packages need a locate-library guard as opposed to a featurep guard? Or which packages need a guard at all?

Yep.  I'd rather we know **why** the guards are in place rather than just do it as a pattern everywhere.   That said, a good pattern to follow is still a good thing (tm).

jvdydev:
One "downside" I noticed with `locate-library` (while experimenting getting the screencast module to work) is that if the library is still present (from previous setups) it will still load.

Assuming a cleaned config dir (no packages installed):
- initial load with `-packages` installs the package (e.g. keycast), `-config` loads configuration
- removing `-packages` from init.el (keycast will no longer be in `package-selected-packages`), keep `-config` in init.el
- subsequent runs will still run the configuration in `-config` (when using `locate-library` as guard, library is still in load-path)

Doesn't matter in the keycast case because users probably also want to remove running the `-config` as well.
However for granular package selection (e.g. removing one out of the completion stack) and cross-module interactions like here, I'm pretty sure this will cause some unintended behavior.

The only work-around (without checking `package-selected-packages`) I see would be to uninstall all unused packages (e.g. using `package-autoremove`)?

jeffbowman:
Yes, that actually makes sense.  If you installed the package, the remove the `-packages` module from the `init.el`, you didn't actually uninstall the package which means it is still  there and on the `load-path`.  This would also be true with anything that was `featurep` and `boundp` (at least within the same session).  If you restart Emacs, `featurep` and `boundp` may be `nil`, but `locate-library` would still find the file on the `load-path`.   So, yes, you'd have to uninstall the package and (maybe) restart Emacs to fully cleanup the `load-path` so `locate-library` doesn't find it. 

Regarding `package-autoremove`, that has caused things to break for me in interesting ways in the past when I chose to run it after installing updates.  It would remove things that I hadn't loaded but still had configured (for example, `python-mode` but I wasn't doing any Python development), sometimes other things.  I'd want to make sure I did some research to understand how that particular function works before suggesting its use or even using it in Crafted Emacs as (for me) it seems like the result has been a bit random - and sometimes completely unexpected.

jvdydev:
Haven't experimented too much with `package-autoremove`, from the docs it would seem to me that would have to be run *before* loading the configuration modules (e.g. right after `package-install-selected-packages`.

Just wanted to point out my findings as it seems there might be some issues with `locate-library` when solving this problem (notably only removing vertico from the install list which may then result in a weird state regarding which vertical mode is active and/or use vertico even if it's not explicitly installed).

jeffbowman:
> Just wanted to point out my findings as it seems there might be some issues with locate-library when solving this problem (notably only removing vertico from the install list which may then result in a weird state regarding which vertical mode is active and/or use vertico even if it's not explicitly installed).

Thanks!  I do appreciate you doing so!  

I'm not sure there is a straight-way forward for handling this situation - ie when a package is initially installed, then only removed from the list of selected packages (but not uninstalled) and then Emacs is restarted.  Yes, it isn't in the list of installed packages, but it is still installed on the system, and therefore found by `locate-library`.  In the case where we intend for it to be installed, unless `vertico-mode` is called then `(featurep 'vertico)` is `nil`.  So the chicken and egg thing commences.  Howe do we know to start the mode?  Well, the thing has to be installed, right?  So, assuming it is installed, we can run `vertico-mode`.  We could try just `(require 'vertico)` and let that fail if it isn't found - but that seems like a bad user experience.  

Cool, so, what if we accidentally install vertico?  We subsequently removed it from the list of installed packages, but all the above still remains true - we can't tell the difference between the user not installing the package, and not wanting the package to be loaded.  Because `vertico-mode` is an autoload function, it returns `nil` on `featurep` (as we found out during the stream that sparked some of this thread).  We could check the list of selected packages, but that fails when the user is installing them from an external source (like pacman, guix, nix, apt, etc).  

The problem, I think, is one of `intent`.  We can't tell what the user **intends** with certain configuration options in the face of installed packages.  This, to me, implies 2 things (non-exclusively)

1. The user must own their configuration enough to know what packages they want installed and uninstall the others.
2. The user must own their configuration enough to know to turn off things they don't want by simply "outgrowing" the Crafted Emacs configuration modules (thus by copy/pasting the parts they want into their own modules or directly in `init.el` and not keeping anything they don't want). 

I think our "best effort" would be to detect if a library is available to load and then assume the user intends to use the package since they include one of our modules.  We might need to cover this in the docs to highlight the fact this situation exists and the way forward is to (possibly) uninstall the "offending" package. 

Make sense? What do you think?

jvdydev:
Makes sense to me, and I agree with the points on intent.

I did see the configuration modules for `ide` and `writing` actually use the `(when (require <pkg> nil :noerror) ...)` "guard" (with essentially the same issue, loads the config even if package is not selected).

I think documenting the behavior is the best we can do here, inform the users what is happening and how to manage (after all, it is intended behavior of Emacs to hold onto installed packages between reboots).

Regarding the guard, I think `locate-library` describes the intent of the code better *unless* we need to `require` the package anyway:
```emacs-lisp
;; (require 'vertico) is in the guard anyway, so we might as well use it as a guard?
(when (require 'vertico nil :noerror)
  ;; vertico conf here ...
  )

;; alternative with locate-library (should be equivalent)
(when (locate-library 'vertico)
  (require 'vertico)
  ;; vertico conf here ...
  )

;; consult does not need a require for the config, so locate-library anyway
(when (locate-library 'consult)
  ;; consult conf here ...
  )
```
I'm open to both versions personally.

jeffbowman:
@sthesing do you want to update the PR or would you rather one of us take over?

jvdydev:
From #312, maybe @lccambiaghi may want to take this?

sthesing:
> @sthesing do you want to update the PR or would you rather one of us take over?

Sorry, I have been away for a while. I think the solution you found makes sense. I'd gladly put the work in, but I must admit I don't quite know which method to apply to which of the packages.
So, if you could tell me which method to use for which package, I'll amend the PR. But if it's less of a hassle to take over, that's fine, too.

jeffbowman:
> Sorry, I have been away for a while...

No worries.  

> So, if you could tell me which method to use for which package...

Not sure I can do that immediately.  Part of the effort on this PR would be figuring that out through testing and reviewing the different packages.  Not a small amount of effort admittedly. That said, using the `require` guard as a default approach and then fixing the places where the guard didn't work as expected would be a good approach. 

Perhaps you and @lccambiaghi  could work together on it? 

And, of course, I and @jvdydev are here to help if needed.

sthesing:
> Part of the effort on this PR would be figuring that out through testing and reviewing the different packages. Not a small amount of effort admittedly. That said, using the `require` guard as a default approach and then fixing the places where the guard didn't work as expected would be a good approach.

Oh, OK, that's fine then. If it's a matter of trying and testing: _that_ I can do. I'll probably be able to do it for the scope of this PR - i.e. `crafted-defaults-config` - over the weekend. For the other config modules, I might need a while. 
Is it a good idea to prepare PRs on a module-by-module basis? I figure that way, anyone who wants to help can chime in without much need for coordination...

jvdydev:
I think a module-by-module basis is a good call, in case we find some unwanted behavior behind the guard, the PR can work around that as needed without holding the other modules back.

We might want to make a tracking issue (similar to #128) with the modules that need help, are resolved and/or if there is already a PR for them.
I'll look through the modules and set one up later (redirecting to here and including the most important bits from this discussion).

sthesing:
- OK, now this module contains `require` guards for everything that does need to be required anyway. 
- The consult config I put behind a `locate-library` guard. 
- I moved the code that turns of fido and icomplete behind the vertico guard. 
  -  I left the `featurep`-check for the default-config module intact, though. So if fido etc. weren't turned on by us, but by the user themself, we won't interfere.

modules/crafted-completion-config.el
```diff
@@ -43,22 +41,32 @@ ARG is the thing being completed in the minibuffer."
     (with-eval-after-load 'evil
       (keymap-set vertico-map "C-j" 'vertico-next)
       (keymap-set vertico-map "C-k" 'vertico-previous)
-      (keymap-set vertico-map "M-h" 'vertico-directory-up))))
+      (keymap-set vertico-map "M-h" 'vertico-directory-up)))
 
+  ;; Turn off the built-in fido-vertical-mode and icomplete-vertical-mode, if
```
jeffbowman, line 46: Do we need something similar in `crafted-defaults-config.el` since we can't guarantee the order in which the modules are loaded by the user?

jeffbowman (commented):
This looks good to me, perhaps needs a little extra corresponding work to make sure everything works as desired (see note on line 46).

modules/crafted-completion-config.el
sthesing: Thanks for the feedback. Good point. I now used the same logic as with the evil related-settings before. Check with `featurep`, but don't act until the respective module has been loaded. That should be the right way to deal with load order issues, right?

jeffbowman:
The change looks good, thanks.  Have we covered this situation?

    (require 'crafted-completion-config)
    (require 'crafted-defaults-config)

I think the other order (where `crafted-defaults-config` is loaded earlier than `crafted-completion-config`) is handled just fine.

sthesing:
> The change looks good, thanks. Have we covered this situation?
> 
> ```
> (require 'crafted-completion-config)
> (require 'crafted-defaults-config)
> ```

No, in fact we haven't. Thanks for pointing that out. Come to think of it, the `featurep` is not only unnecessary in general, but also a problem here. If we just do:
```elisp
(with-eval-after-load 'crafted-defaults-config
    (fido-mode -1)
    (fido-vertical-mode -1)
    (icomplete-mode -1)
    (icomplete-vertical-mode -1)))
```
fido and icomplete are turned off if vertico is loaded, regardless, of the load order. I tested it for both load orders and also for not using `crafted-defaults-config` altogether. And they all work as expected and without errors.
If I'm not overlooking anything, I'd make that change, OK?

jeffbowman:
Yes, that change looks reasonable.  Thanks for the investigation, development and testing!

sthesing:
Done. Thanks for all the helpful feedback!

jeffbowman:
Closing this and marking #313 complete for the `crafted-completion` module.

Thanks @sthesing for the work on closing this PR.  Please feel free to review other modules (see #313) and add PRs for those as well.

monora:
FYI: I just tested the new version of the completion module and run into an error with vertico-directory (located in the extenstion folder of vertico). I am using straight as package manager (see #315). Therefor the extensions are not on the load-path. The solution is described in https://github.com/radian-software/straight.el/issues/819. In my completion-config I fixed it like so:

```emacs-lisp
;; Before requiring crafted-completion, we must load the vertico extension.
(straight-use-package '( vertico :files (:defaults "extensions/*")
                         :includes (vertico-directory
                                    )))
(require 'crafted-completion-config)
;; my additional completion config follows
```

Should we put these notes somewhere in the doc? Although straight is not supported...

jeffbowman:
Yep, this kind of thing happens all the time with `straight.el` users.  When the package is built for the ELPAs, things get flattened, although there is (has been) discussion about changing that.  `package-vc` would likely have the same problem, as would `borg` and possibly `quelpa`, each with their own different approaches to solve the same problem. 

We assume if a user prefers a different package manager, they are able to handle these situations or find the appropriate solution on their own.  If we wanted to be generous we _could_ provide a "Tips and Tricks" area for things like this.  If we turn on the wiki, that might be a better place for that kind of content. 

@jvdydev thoughts?

jvdydev:
agreed, probably best placed in a Tips & Tricks section (or as a note on using other package managers in the existing different package manager section).

Also probably should move this conversation out to an issue so it is easier to find for other people using those package managers.

Issue #295, opened 2023-07-04
[craftedv2beta] org-mode-hook may fail because of void variable


jvdydev:
While experimenting some more with the `crafted-org` module, I found that Emacs would not finish running `org-mode-hook` and presented the following error in the minibuffer:
```
File mode specification error: (void-variable electric-pair-inhibit-predicate)
```

On subsequent manual activation of `org-mode`:
```
Symbol's value as variable is void: electric-pair-inhibit-predicate
```

Interacting with `electric-pair-mode` in some way (turning on/off) fixed the error from what I can tell.

I found a temporary fix by checking if the variable is bound before setting it locally (reformatted):
```emacs-lisp
(add-hook 'org-mode-hook
          (lambda ()
            (when (boundp 'electric-pair-inhibit-predicate)
              (setq-local electric-pair-inhibit-predicate
                          `(lambda (c)
                             (if (char-equal c ?<)
                                 t
                               (,electric-pair-inhibit-predicate c)))))))
```

Might also be worth pulling it out into a separate function anyway (see #280).

jeffbowman:
> I found a temporary fix by checking if the variable is bound before setting it locally (reformatted):

Do you consider this to be a temporary fix because:
>   This variable may be risky if used as a file-local variable.

I'd have to look into why it would be considered risky, but your solution seems reasonable enough to me for the moment. 

What would you consider to be a more permanent fix?  

I've not run into this problem, but that is likely because I turn on `electric-pair-mode` before I configure org-mode.  Also, I wonder if we use `customize-set-variable` would the error go away the next time Emacs starts and runs this configuration(?) because the `custom-file` is loaded first / early in `init.el` (assuming the user follows our suggestion on structure for that file).

jvdydev:
I mainly considered it a temporary fix cause the entire thing should be a named function (#280) and it was more of a "how do I get this working at all" solution without too much thought.
I haven't looked into the issue further than the fix above either.

The fix up top would fail for a situation where `electric-pair-mode` is activated while already in org-mode (if `electric-pair-mode` was not activated before).
However I think that's a *very* niche issue and is fixed by just re-entering org-mode.

Regarding `customize-set-variable`, wouldn't that be a global set (meaning it would apply to non-`org-mode` buffers)?
A quick'n'dirty test gave me the same error, but that would be something to look into more closely.

[...]

PR #312, opened 2023-07-27
Use "locate-library" to check if a package is installed

lccambiaghi:
Hi,

Today I finally decided to migrate to crafted-emacs v2. Thank you so much for your work, I really like the new way of configuring it!!

Reading the guide it seemed very straightforward to write my new `init.el`:
- Require all the "crafted-xxx-package" modules I am interested in
- Install the packages with `(package-install-selected-packages :noconfirm)`
- Require all the "crafted-xxx-config" modules I am interested

However, after doing so, I saw that the packages (e.g. `vertico` and `marginalia`) were not being activated/configured. After spending one hour on the issue, I realized it is beacuse of the `(when (featurep 'vertico) ... )` checks.

I then watched the System Crafters stream and I saw David Wilson struggle with my same issue. As explained in the video, `featurep` returns `t` only if the package has already been `require`'d, which I think is not the intended behavior.

In the comments of the video a kind user recommended to use `locate-library` to check if a package is installed. This PR changes the checks to use `locate-library` for all non-deprecated modules in the `modules` folder.


jvdydev:
Thank you for your PR and for testing the beta.

While we appreciate that you went out of your way to fix a lot of current issues, there are some things I want to point out (nothing personal, just things I noticed while trying to understand your PR):

Your PR names only one of the (at least) 5 things it solves. There is also generally a lot of changes in a lot of different places at once, which would preferably be split out (within reason).

Your fix for `featurep` -> `locate-library` doesn't follow the result of the discussion (using `require` as a guard where appropriate, see #294). While there is no one true solution, we'd rather want to discuss the way we want to do a thing and then figuring out a solution that is maintainable long-term (effectively to establish a coding style/convention). The commit also contains a bug, passing the symbol for `corfu-terminal` instead of the string (like everywhere else in the commit).

Your fix for conflicting narrowing frameworks is essentially the same fix as originally proposed in #294. This means it suffers from the same issues as pointed out there (doesn't account for manual package selection, different loading orders, etc.).

I appreciate the `with-eval-after-load` solution for #295. It's a better solution than my last proposal.

I don't quite understand the "Don't install all grammars by default" commit. Your change only removes the `with-eval-after-load` and still installs all the grammars, just at Emacs startup, not when `treesit-auto` is loaded. So the commit message and the change don't line up. I'm also not sure what this fixes (please explain).

Regarding the last commit, it would be nice if you could provide more explanation for what it solves. I am not aware of any open issues about this, so it's hard to evaluate why this specifically is the appropriate solution to the problem.

lccambiaghi:
Hi Judy, I am sorry about the confusion. I have pushed by mistake to the `craftedv2beta` of my fork and those commits ended up on this PR against my will. Please only consider the first commit, relevant for this PR. I also read just now the dicussion in #294, sorry I was not aware of the alternatives proposed.

Feel free to close this PR until you can reach an agreement on what is the best way to solve this issue. 

My intentions were good: I thought I could be of help to another user like me who would migrate to v2 just to see their packages not being installed. I think it is a very critical bug and I would personally rather fix it (e.g. by using `locate-library`) and then potentially change again considering the other alternatives. Unfortunately the current version does not work as intended.

jvdydev:
No worries, looking back at the commit history, I suspected this might have been the case :smile: .

I agree that the `locate-library` situation should be fixed sooner rather than later so we can find all the issues hidden behind the guards that don't work.

Would appreciate if we can move the discussion if you want to take over this to #294 and move from there (closing this)?

Issue #280, opened 2023-06-06
(style) Make named functions for hook lambdas


japhir:
the style guide says to not use lambdas for hooks
https://github.com/bbatsov/emacs-lisp-style-guide#functions

I see lambda hooks in crafted-writing-config and in crafted-org-config.

PR #314, opened 2023-07-29
[craftedv2beta] Fix electric-pair-inhibit-predicate in crafted-org

jvdydev:
Addresses #295 (and partially #280) by implementing a fix to check modes are enabled.

Added to both mode-hooks to ensure the order of activation for `org-mode` and `electric-pair-mode` is irrelevant for producing the effect (setting the buffer-local variable).

Documentation is updated (info buffer regenerated).

Ready for review/merge.


jeffbowman:
Looks good to me.

Issue #313, opened 2023-07-29
[craftedv2beta] Broken configuration guards (Tracking Issue)


jvdydev:
Tracking issue for the broken guard issue in craftedv2beta.

Affected modules (contain `featurep`):
- [x] crafted-completion (PR #294)
- [x] crafted-defaults (PR #317)
- [x] crafted-evil (PR #319)
- [x] crafted-ide (PR #320)
- [x] crafted-lisp (PR #328)
- [x] crafted-org (PR #322)
- [x] crafted-osx (PR #343)
- [x] crafted-screencast (PR #304)
- [x] crafted-ui (PR #318)
- [x] crafted-writing (PR #329)

Since the guards don't work, there might be more bugs below, therefor it would be nice to deal with these on a module-by-module basis (as suggested by @sthesing).

The proposed fixed guard patterns we want to apply (see #294 for the discussion).

```emacs-lisp
;; (require 'vertico) is in the guard anyway, so we might as well use it as a guard?
(when (require 'vertico nil :noerror)
    ;; vertico conf here ...
    )

;; consult does not need a require for the config, so locate-library anyway
(when (locate-library 'consult)
    ;; consult conf here ...
    )

;; maybe (as needed)
(when (or (featurep 'vertico)
          (require 'vertico nil :noerrror))  ; whichever guard as appropriate
    ;; vertico conf here
    )
```

Quote from #294 by @jeffbowman regarding how to apply the pattern above:

```markdown
[I]f we need to require something anyway, lets use the require guard, 
otherwise lets consider using the `locate-library` guard. 
Mindfully, we can add `featurep` with `or` if that makes sense 
(although I expect there will be very few examples where that would be useful).
```

Further discussion and/or contributions welcome :smile:

monora:
Could we provide a macro for the two use-cases? Perhaps like this:

```emacs-lisp
(defmacro crafted-when-package-installed-do (package use-require &rest body)
  `(when (if ,use-require
             (require ,package nil :noerror)
           (locate-library (symbol-name ,package)))
           ,@body))
```
Usage:
```emacs-lisp
(crafted-when-package-installed-do 'vertico t
  ;; vertico conf)

(crafted-when-package-installed-do 'consult nil
  ;; vertico conf)
```

Then we could easily find the usage of the pattern.

jeffbowman:
[...]

PR #319, opened 2023-08-07
[craftedv2beta] `locate-library`-guard for `crafted-evil-config`

sthesing:
As discussed in #313


jeffbowman:
Looks good, thanks!

PR #322, opened 2023-08-08
[craftedv2beta] `locate-library`-guard for `crafted-org-config`

sthesing:
As discussed in #313 .

I had a thought, though: When it's only a question of whether a mode is available or not (as in this case), wouldn't `fboundp` be more efficient? And the more canonical way to do this?

```elisp
(when (fboundp 'org-appear-mode)
  (add-hook 'org-mode-hook 'org-appear-mode))
```

I like the simplicity of using a `require`-guard if we require anyway and using `locate-library` for anything else, but a case like this made me wonder.

Any thoughts?

### DIFF
--- a/modules/crafted-org-config.el
+++ b/modules/crafted-org-config.el
@@ -23,7 +23,7 @@
 
 ;; Hide markup markers
 (customize-set-variable 'org-hide-emphasis-markers t)
-(when (featurep 'org-appear)
+(when (locate-library "org-appear")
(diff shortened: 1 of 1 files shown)


jeffbowman:
`fboundp` is useful for checking if a function from a built-in package or an autoload function is bound. In this case, if the `org-appear` package is installed and activated by the package manager (ie it is on the `load-path` and the autoloads have been loaded), then `fboundp` is a good solution.

jeffbowman:
Looks good, thanks!

PR #323, opened 2023-08-08
Improve documentation to bootstrap straight package manager

### LINKED ISSUE #315: [craftedv2beta] Improve documentation for using straight.el
I finally managed to setup `craftedv2beta` to use `straight` instead of the builtin package manager. The doc alone _(info "(crafted-emacs)Using alternate package managers"')_ was not enough to get started. Here is my setup:

early-init.el:

```emacs-lisp
(setq crafted-emacs-home "~/crafted-emacs-v2"
      crafted-bootstrap-directory (expand-file-name "bootstrap" crafted-emacs-home)
      crafted-config-path user-emacs-directory)
(load (expand-file-name "crafted-package" crafted-bootstrap-directory))
(setq crafted-package-system 'straight)
(crafted-package-bootstrap)

;; Adds crafted-emacs modules to the `load-path', sets up a module
;; writing template, sets the `crafted-emacs-home' variable, unless not already set.
;; We set it in `early-init' explicitly.
(load (expand-file-name "modules/crafted-init-config" crafted-emacs-home))

;; Now we can require this and configure `straight' as explained in `(info "(crafted-emacs)Using alternate package managers"'
(require 'crafted-package-config)

(setq crafted-package-installer #'straight-use-package)
(setq crafted-package-installed-predicate #'straight--installed-p)

;; Should we have an `crafted-early-init-straight-config' containing the code in this file ?
;; (load "~/crafted-emacs-v2/modules/crafted-early-init-config")
```

init.el:

```emacs-lisp
;;; Initial phase.

;; Load the custom file if it exists.  Among other settings, this will
;; have the list `package-selected-packages', so we need to load that
;; before adding more packages.  The value of the `custom-file'
;; variable must be set appropriately, by default the value is nil.
;; This can be done here, or in the early-init.el file.
(setq custom-file (expand-file-name "custom.el" user-emacs-directory))
(when (and custom-file
           (file-exists-p custom-file))
  (load custom-file nil 'nomessage))

;; Adds crafted-emacs modules to the `load-path', sets up a module
;; writing template, sets the `crafted-emacs-home' variable.
;; (load (expand-file-name "modules/crafted-init-config" crafted-emacs-home))
;; already done in early-init.el, to configure package-system.

;;; Packages phase

;; Collect list of packages to install.

;; Install the packages listed in the `package-selected-packages' list. CHANGE: Must use
;; `crafted-package-install-selected-packages' instead of
;; `package-install-selected-packages'
(crafted-package-install-selected-packages)

;; Rest same as in example init.el
```

Is this the intended way to use _straight_ as package manager? If so, i would recommend to provide these two files as additional *init.el examples in the _examples_ directory and refer to them in the docs.

monora:
fixes #315

### DIFF
--- a/docs/crafted-emacs.org
+++ b/docs/crafted-emacs.org
@@ -127,14 +127,20 @@ for the code examples.
   that, the ~crafted-package-config~ module provides a variable to set
   and a function to call.
 
-  Here is an example using ~straight-use-package~ as the tool to
-  install packages:
+  Here is an example using ~straight-use-package~ as the tool to install
+  packages. The following lines must be called before any package is installed
+  (preferable in ~early-init.el~):
 
   #+begin_src emacs-lisp
+    (load (expand-file-name "modules/crafted-package-config" crafted-emacs-home))
+ 
     (setq crafted-package-installer #'straight-use-package)
     (setq crafted-package-installed-predicate #'straight--installed-p)
-    (crafted-package-install-package-list)
   #+end_src
+
+  In your ~init.el~ you have to call ~crafted-package-install-selected-packages~
+  instead of ~package-install-selected-packages~, after adding your packages to
+  ~package-selected-packages~.
 
   Note: This usage is simple in the sense full recipes are not used
   when using @@texinfo:@emph{Crafted Emacs}@@ package bundle modules.
@@ -142,6 +148,18 @@ for the code examples.
 
   #+begin_src emacs-lisp
     (straight-use-package 'corfu)
+  #+end_src
+
+  You find a full example [[file:../examples/crafted-early-init-straight.el][crafted-early-init-straight.el]] to bootstrap straight
+  in the examples directory. Merge it into your ~early-init.el~ or copy it to your
+  ~custom-modules~ folder. In the later case ~early-init.el~ would look like
+  this:
+
+  #+begin_src emacs-lisp
+  (setq crafted-emacs-home "~/crafted-emacs")
+
+  (load (expand-file-name "custom-modules/crafted-early-init-straight"
+                          user-emacs-directory))
   #+end_src
 
   It is not required to use the @@texinfo:@emph{Crafted Emacs}@@
--- a/examples/crafted-early-init-straight.el
+++ b/examples/crafted-early-init-straight.el
@@ -1,0 +1,42 @@
+;;; crafted-early-init-straight.el --- Bootstrap straight.el  -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2023
+;; SPDX-License-Identifier: MIT
+
+;; Author: System Crafters Community
+
+;;; Commentary:
+
+;; Code to bootstrap straight package manager
+;;
+;; Load it in early-init.el like this:
+
+;; (setq crafted-emacs-home "~/crafted-emacs")
+;; (load (expand-file-name "custom-modules/crafted-early-init-straight"
+;;                         user-emacs-directory))
+
+;;; Code:
+
+(load (expand-file-name "modules/crafted-package-config" crafted-emacs-home))
+
+;; See https://github.com/radian-software/straight.el#getting-started
+(defvar bootstrap-version)
+(let ((bootstrap-file
+       (expand-file-name "straight/repos/straight.el/bootstrap.el" user-emacs-directory))
+      (bootstrap-version 6))
+  (unless (file-exists-p bootstrap-file)
+    (with-current-buffer
+        (url-retrieve-synchronously
+         "https://raw.githubusercontent.com/radian-software/straight.el/develop/install.el"
+         'silent 'inhibit-cookies)
+      (goto-char (point-max))
+      (eval-print-last-sexp)))
+  (load bootstrap-file nil 'nomessage))
+
+;; Configure crafted-emacs to use straight as package manager.
+;; See `(info "(crafted-emacs)Using alternate package managers")'
+(setq crafted-package-system 'straight)
+(setq crafted-package-installer #'straight-use-package)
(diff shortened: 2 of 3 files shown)


jeffbowman:
The info file needs to be regenerated.  

The `crafted-init-config` module sets the `crafted-emacs-home` variable, it does so by deducing the path where the `crafted-emacs` project code lives, as long as the value hasn't already been set.  So, it is safe to set in the `early-init.el` file, but I wonder if we might want to use a similar method. 

I'm not sold on copying the example to the `custom-modules` path and then loading it from there when it could easily be the user's `early-init.el` file by itself.  Doing so would mean the `crafted-emacs-home` value doesn't need to be set there, so less confusion on when/where/how to set that variable.  Additionally, if the `straight.el` maintainer devises a new method to bootstrap `straight`, then the example becomes a source of bugs, whereas depending on the user to own their own configuration means we don't need to track that process.  Technically, it is just **an example**, but in the face of such a change, someone will file a bug report that the example is wrong.  

On the other side of that coin, having a working example like this would be valuable and helpful to many, and the risk of changes isn't actually that high... sooo... I'm mostly in favor of merging this PR, but perhaps @jvdydev will share some thoughts as well.

jvdydev:
I think the idea of providing an example file for straight.el is generally a good idea (long-term I want to expand the examples section, but that should do for now).
Updating a broken example because of bootstrapping changes feels much less overhead than a module (maybe with a nice reminder that we don't officially support other package managers).

I agree on the copying an example and using it as a module (being equally not sold on it).
I think of the examples as a "copy in as my early-init/init and play around/get started"-type deal compared to the modules which are configuration blocks that are essential to the day-to-day usage.

I think personally, the only change on the docs should be maybe adding how to load the `crafted-package-config` (potentially setting `crafted-emacs-home` in the process, I'm open for either way on that one).
Otherwise the docs section above the "Here is an example using straight.el" is quite verbose on where those variables come from (in my opinion).

docs/crafted-emacs.org
```diff
@@ -131,11 +131,28 @@ for the code examples.
   install packages:
 
   #+begin_src emacs-lisp
```
jvdydev, line 133: Maybe adding how to load `crafted-package-config` here would be nice

docs/crafted-emacs.org
```diff
@@ -131,11 +131,28 @@ for the code examples.
...
-    (setq crafted-package-installer #'straight-use-package)
-    (setq crafted-package-installed-predicate #'straight--installed-p)
-    (crafted-package-install-package-list)
+  (setq crafted-package-installer #'straight-use-package)
+  (setq crafted-package-installed-predicate #'straight--installed-p)
+  (crafted-package-install-package-list)
   #+end_src
 
+  You can find an example to bootstrap straight in the examples directory:
```
jvdydev, line 139: I'd personally keep the "You can find an example" part, but remove the copying to custom-modules block.

examples/crafted-early-init-straight.el
```diff
@@ -0,0 +1,42 @@
...
+;; Copyright (C) 2023
+;; SPDX-License-Identifier: MIT
+
+;; Author: System Crafters Community
+
+;;; Commentary:
+
+;; Code to bootstrap straight package manager
+;;
+;; Copy it to your custom-modules folder and load it in early-init.el like this:
```
jvdydev, line 12: Same here on the copying to custom-modules

jvdydev (commented):
Basically what my previous comment said, just with local comments, see above for some of the reasoning.

monora:
With 267b5c4 I hope I resolved all comments. I reworded the doc again a little. Please have a look.

docs/crafted-emacs.org
```diff
@@ -144,6 +150,23 @@ for the code examples.
     (straight-use-package 'corfu)
   #+end_src
 
+  You can find a full example to bootstrap straight in the examples directory:
+
+  #+begin_src emacs-lisp
+    (find-file "../examples/crafted-early-init-straight.el")
```
jeffbowman, line 156:
```suggestion
    [[file:../examples/crafted-early-init-straight.el][crafted-early-init-straight.el]]
```

This should be a link as this document could be viewed online or in info.  (and therefore don't need the src block tags either. )

docs/crafted-emacs.org
```diff
@@ -144,6 +150,23 @@ for the code examples.
...
+  #+end_src
+
+  Assuming you copied it to your ~custom-modules~ folder, you can load it in
```
jeffbowman, line 159: I might consider this an alternate approach.  The example file could be copied (or merged) with/as the user's `early-init.el` file

jeffbowman (changes requested):
A few more changes.  This is good work and I really appreciate the effort you are putting into it.  Thanks for being willing to work with us on this PR! 

cc: @jvdydev

jeffbowman:
> Do you squash, when you merge?

I usually just rebase, but you are welcome to squash if you wish.

docs/crafted-emacs.org
```diff
@@ -144,6 +150,23 @@ for the code examples.
...
+  your ~early-init~.el like this:
+
+  #+begin_src emacs-lisp
+  (setq crafted-emacs-home "~/crafted-emacs")
+
+  (load (expand-file-name "modules/crafted-package-config" crafted-emacs-home))
```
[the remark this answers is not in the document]
monora: Done with fa8e2aed

monora:
> > Do you squash, when you merge?
> 
> I usually just rebase, but you are welcome to squash if you wish.

OK, than I would squash manually, so that you can merge if you are satisfied with my last changes.

jeffbowman (approved):

monora:
Squashed to one commit. Ready to merge.

PR #325, opened 2023-08-12
string-match-guards for crafted-init-config

UristMcLisper:
The functionality for ignoring the auto-insert on the custom file may run into an issue 
when the `custom-file` and/or `file-buffer-name` values are `nil`.

This essentially breaks the `auto-insert` function for non-file buffers  (and all buffers 
if `custom-file` is `nil`).

### DIFF
--- a/modules/crafted-init-config.el
+++ b/modules/crafted-init-config.el
@@ -82,7 +82,8 @@ explicitly."))
     "Apply ORIG-AUTO-INSERT only when the file is not the
 `custom-file' to avoid confusion when that file doesn't exist on
 startup."
-    (if (string-match (file-name-nondirectory custom-file) buffer-file-name)
+    (if (and custom-file buffer-file-name
+             (string-match (file-name-nondirectory custom-file) buffer-file-name))
         (message "Skipping auto-insert for %s" custom-file)
       (apply orig-auto-insert args)))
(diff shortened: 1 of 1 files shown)


jvdydev:
Is the commit author just an error in a git config (not the same username)?

Additionally, it would be nice to add some comment on *why* this is changed. 
It's much easier for us to figure out what you want changed (especially if there is no issue attached).
Doesn't have to be long, something like:
```markdown
The functionality for ignoring the auto-insert on the custom file may run into an issue 
when the `custom-file` and/or `file-buffer-name` values are `nil`.

This essentially breaks the `auto-insert` function for non-file buffers  (and all buffers 
if `custom-file` is `nil`).
```

UristMcLisper:
> Is the commit author just an error in a git config (not the same username)?
That's intentional. So that guix people can assign copyright based on git name ;)

UristMcLisper:
1. Follow getting started from README.org
2. Emacs starts with failure and any find-file displays buffer only on second try. 
3. I added check for custom-file
4. Same thing happened later when I tried to open .epub file usinv nov.el package. 
5. I added second check

Nothing to document I think

jeffbowman:
Tried those steps, everything worked perfectly for me.  I'll merge this anyway, it doesn't do any harm, and if it helps someone else, then we still win. 

Thanks again!

Issue #324, opened 2023-08-10
[craftedv2beta] Use `customize-save-variable` vs. `customize-set-variable`?


sthesing:
Since v2beta, the approach to customizations is to push customized values to the custom file (as explained by @jeffbowman in the [respective blog post](https://jeffbowman.writeas.com/crafted-emacs-v2-beta)). 
Currently, this is accomplished by using `customize-set-variable` and running `customize-save-customized` after the init is loaded. Minimal example init.el:
```elisp
(setq custom-file (expand-file-name "custom.el" user-emacs-directory))
(customize-set-variable 'load-prefer-newer t)
(add-hook 'after-init-hook #'customize-save-customized)
```
The same could be achieved by using `customize-save-variable` instead:
```elisp
(setq custom-file (expand-file-name "custom.el" user-emacs-directory))
(customize-save-variable 'load-prefer-newer t)
```
That would remove the need for the running `customize-save-customized` in the `after-init-hook`.

Are there any reasons to prefer `customize-set-variable` over `customize-save-variable`? 
In the [old readme](https://github.com/SystemCrafters/crafted-emacs/blob/craftedv2beta/README-old.org), there is a section about values shown as "SET for current session only" in the Customization UI, which I frankly don't understand. Does that have something to do with this?

# Related issue

I'm asking about this because I observed a somewhat puzzling behaviour with e.g. `consult-theme` in crafted emacs. 
Using it to select another theme results in an error:
```
Error in post-command-hook (consult--preview-post-command-hook): (invalid-function '(("gnu" . 99) ("nongnu" . 80) ("stable" . 70) ("melpa" . 0))
```
or (if `crafted-early-init-config` is not used):
```
Error in post-command-hook (consult--preview-post-command-hook): (void-function t)
```
Apparently, consult's `consult--preview-post-command-hook` tries to run a customized variable as a function. It has to do with the `after-init-hook` that somehow leaks into consult's hook.

I tracked the issue down to the function `consult--preview-append-local-pch` in consult.el, which mentions a [bug in Emacs](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=46326#29) that apparently interferes with cleaning up `hook--depth-alist` and thus can cause hooks leaking into each other.  is not cleaned up properly, see also minad/consult#193.

I'm not quite sure whether it is something that Daniel could easily workaround in consult, but - long story short:
[...]

PR #330, opened 2023-08-16
Replaced `customize-save-variable` with `customize-set-variable`

sthesing:
As discussed in #324, `customize-set-variable` is preferred over `customize-save-variable`.
Replaced all left over occurrences of `customize-save-variable`.


jeffbowman:
Looks good, thank you!

PR #331, opened 2023-08-17
[craftedv2beta] Add documentation for `crafted-completion`

sthesing:
I adapted the docs for `crafted-completion` that I originally wrote for v1. Docs for Corfu and Cape included. 

I hope the original approach with explaining the packages step-by-step with screenshots is still alright.


jvdydev:
Thank you, this already looks really good.

I wonder if instead of having a "Documentation" heading, instead having a heading for each package (given the in-depth explanations).
I assume you tried to adapt to what the other docs already have (which I appreciate), however I think this would be a case where breaking things out may help.

Also some very small polishing things like typos or accidental characters (not meant negatively, just something I noticed. I would be doomed without spell checkers).

sthesing:
I have introduced subheadings now, is this what you head in mind? 
I'm doomed without spell checking, too, and even with the help of it, I found only one typo. If you spotted more, I'd be grateful for pointing me towards them.
Thanks for your feedback!

jeffbowman:
Screenshots are nice to tell a story, but we only have them for this one configuration.  It's harder to imagine screenshots for all other modules, but in the interest of consistency, are there any other modules we should consider a similar approach?

sthesing:
> Screenshots are nice to tell a story, but we only have them for this one configuration. It's harder to imagine screenshots for all other modules, but in the interest of consistency, are there any other modules we should consider a similar approach?

I did consider them while drafting the documentation for `crafted-ui`, too (PR coming soon). My first thought was that `crafted-completion` and `crafted-ui`  affect how the interface _looks_, so I thought both might benefit from screenshots.
However, the look aspect is not so much the topic here, it's more a matter of positioning: where does Emacs put which information presented to the user?
In that, `crafted-completion` is rather special, so I do think screenshots are helpful, here.
But, as I wrote back in #128 and #155, I tried to write the text so that it works without screenshots, too. (Not sure if I succeeded, though...)

jeffbowman:
Yep, and I'm a fan of the screenshots, just wanted to ask if there were other places they made sense as well.

sthesing:
> just wanted to ask if there were other places they made sense as well.

I've been thinking about this a bit, and have toyed with the idea whether it might be worthwhile to add screenshots to the documentation of `crafted-screencast` or `crafted-workspaces`. But just as with `crafted-ui`, I feel that in both cases the brevity of pure text is preferable to using screenshots. What do you think?

Do you have any module in mind where it might be worth looking into the benefit of screenshots?

P.S.: I merged the docs for crafted-writing and regenerated the info file, so this should be mergable now.

jeffbowman:
Still seeing merge conflicts on this PR, which is the only reason it has not been merged.

jeffbowman:
Sure, give it a shot!

sthesing:
> Still seeing merge conflicts on this PR, which is the only reason it has not been merged.

Strange, I also don't see conflicts.

Thanks, for trying @jvdydev . If I did something wrong, give me a holler...

jvdydev:
I think it was confused because the commits are older than some of the other PRs that already got merged (hence there was a lot of rebasing different commits to keep the timeline accurate)?

Anyway, seems like it worked

PR #332, opened 2023-08-21
[craftedv2beta] Add option to opt out of saving customizations every session.

sthesing:
As we discussed in #324, the calls to `customize-save-customized` and `package--save-selected-packages` in the `after-init-hook` should be the default behaviour. This adds the possibility for users to opt out of that without having to maintain a custom branch of Crafted Emacs.

It 
- introduces a new customization group "Crafted Init"
- introduces two new variables in that group:
  1. `crafted-init-auto-save-customized`
  2. `crafted-init-auto-save-selected-packages`

By default, they are non-nil. Setting them to nil will prevent `customize-save-customized` and `package--save-selected-packages` to be added to `after-init-hook` respectively.

### DIFF
--- a/modules/crafted-init-config.el
+++ b/modules/crafted-init-config.el
@@ -18,6 +18,27 @@
 ;; including the template below used for writing Crafted Emacs
 ;; modules.
 
+(defgroup crafted-init '()
+  "Initialization configuration for Crafted Emacs"
+  :tag "Crafted Init"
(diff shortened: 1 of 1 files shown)


jeffbowman (approved):
This looks good to me.  @jvdydev any concerns?

jvdydev:
Looks good to me too.

PR #333, opened 2023-08-21
[craftedv2beta] Only check for updates if the user loads `crafted-updates-config`

sthesing:
I noticed that when users load `crafted-startup-config`, it doesn't matter whether they also load `crafted-updates-config`, because it gets loaded by the former.

This fixes that behaviour by only running the update related code in `crafted-startup-config` if (and after) `crafted-updates-config` is loaded.

### DIFF
--- a/modules/crafted-startup-config.el
+++ b/modules/crafted-startup-config.el
@@ -1,6 +1,6 @@
 ;;; crafted-startup-config.el --- Crafted Emacs splash screen on startup  -*- lexical-binding: t; -*-
 
-;; Copyright (C) 2022
+;; Copyright (C) 2023
 ;; SPDX-License-Identifier: MIT
 
 ;; Author: System Crafters Community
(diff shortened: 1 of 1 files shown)


jeffbowman:
Yes, the intention is to have the startup screen show the existence of updates so the user can keep their Crafted Emacs up-to-date (if they wish).  This is an interesting PR as it decouples the two modules, and one of our goals is to have the modules be as independent of each other as possible - ie, we prefer to _not_ have them depend on each other. 

I would wonder if maybe we should add `###autoload` markers to a few forms in `crafted-update-config` which would actually facilitate approximately the same thing - it would break the coupling of the explicit `require`, but still allow the checks for updates without the guards.  Even adding the `###autoload` markers would not invalidate the guards in this case and would otherwise not have any effect on this PR, except to potentially add those markers - but that can be a separate PR (and probably should be). 

@jvdydev any thoughts on this?  I'm considering merging as-is and then opening an issue for the autoload suggestion here.

jvdydev:
ye, splitting out the `###autoload` into a separate issue makes sense.
Code looks good.

PR #335, opened 2023-08-23
[craftedv2beta] Simplify path construction for auto-insert of CE header

jvdydev:
Small potential bug I found while reading into the skeleton discussion.

Replacing
```emacs-lisp
(concat (expand-file-name crafted-emacs-home) "modules/crafted-.*\\.el")
```
with
```emacs-lisp
(expand-file-name "modules/crafted-.*\\.el" crafted-emacs-home)
```

Just a bit "cleaner".
Not sure if it's technically a bug-fix, but if `crafted-emacs-home` had no trailing "/", the previous version would not account for it and produce an invalid path.


jeffbowman (approved):
Nice catch.

PR #340, opened 2023-09-01
[craftedv2beta] Module documentation: `crafted-updates`

sthesing:
Next in line from #128.

### DIFF
--- a/docs/crafted-updates.org
+++ b/docs/crafted-updates.org
@@ -1,0 +1,93 @@
+* Crafted Emacs Updates Module
+
+** Installation
+
+To use this module, simply require them in your =init.el= at the appropriate
+points.
+
+#+begin_src emacs-lisp
+;; No additional packages are installed by this module.
+
+;; install the packages
+(package-install-selected-packages :noconfirm)
+
+;; Load crafted-updates configuration
+(require 'crafted-updates-config)
+#+end_src
+
+** Description
+
+The =crafted-updates= module provides functionality to check for updates of
+Crafted Emacs. To do so, it relies on [[https://git-scm.com/][git]]. It will only work if you cloned the
+Crafted Emacs [[https://github.com/SystemCrafters/crafted-emacs][github repository]] to your machine.
+
+If you installed Crafted Emacs by other means, it's probably a good idea not to
+load this module, but take care of updates manually.
+
+*** Usage and Customization
+
+You can use the functionality provided by this module in three ways:
+
+1. On startup
+2. Regularly
+3. Manually
+
+**** On Startup
+
+This is probably the easiest way. Just load the module =crafted-startup-config=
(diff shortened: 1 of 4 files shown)


modules/crafted-updates-config.el
```diff
@@ -10,16 +10,9 @@
...
-;; don't know where the project was cloned to.  Formerly it was
-;; `user-emacs-directory', however, now we allow the user to clone
-;; this project wherever, add it (ostensibly the modules folder) to
-;; the `load-path' and then build their configuration however they see
-;; fit.  Unless we set a variable like `crafted-emacs-home' there is
-;; no good way to know where this project is cloned.
-
-;; UPDATE: [2023-02-09 Thu] `crafted-emacs-home' is now set on load,
-;; this can be updated to use that value for checking on updates from
+;; This uses `crafted-emacs-home', which is set on load by if it hasn't been
```
jvdydev, line 13: I might be reading this wrong, but there is a "by" too much?

```
This uses `crafted-emacs-home', which is set on load if it hasn't been 
set by the user or `crafted-init-config'.
```

docs/crafted-updates.org
```diff
@@ -0,0 +1,93 @@
...
+Crafted Emacs [[https://github.com/SystemCrafters/crafted-emacs][github repository]] to your machine.
+
+If you installed Crafted Emacs by other means, it's probably a good idea not to
+load this module, but take care of updates manually.
+
+*** Usage and Customization
+
+You can use the functionality provided by this module in three ways:
+
+1. On startup
```
jvdydev, line 31: I'm not sure on this mini "table of contents" when the content is directly following.

Especially in info-buffer, it looks a bit funny given 4-star headings are turned into numbered entries:
```
1. On startup
2. Regularly
3. Manually

1. On Startup
...
```
I'd personally remove it.

jvdydev (commented):
Just a few things to discuss (nothing serious), otherwise looks good.

modules/crafted-updates-config.el
sthesing: Oh yes, that's right. Good eye, thank you!

docs/crafted-updates.org
sthesing: Good point. I agree.
Would you just remove it or turn it into a regular introductory sentence? "You can use the functionality provided by this module in three ways: on startup, regulary and manually."

jvdydev: I think putting it into the introductory sentence makes sense.

sthesing: > I think putting it into the introductory sentence makes sense.

Think so, too. Amended it. Thanks for the feedback!

jvdydev:
Regenerating info buffer, otherwise consider this approved from my end

jvdydev (approved):

PR #342, opened 2023-09-07
Fix config guard for treesit-auto

jeffbowman:
treesit-auto was using locate-library, but should have been using the require guard instead.


sthesing (approved):
Looks good to me.

PR #343, opened 2023-09-08
Update osx module, remove bad assumptions and failing code

jeffbowman:
Removed an assumption a dark theme was being used, also there seemed to be some attempt at setting the name of the selected frame to nil, it didn't work and not sure that makes any sense anyway.

### DIFF
--- a/modules/crafted-osx-config.el
+++ b/modules/crafted-osx-config.el
@@ -16,10 +16,8 @@
 (defun crafted-osx-transparent-titlebar ()
   "Set the titlebar to be transparent."
   (interactive)
-  (customize-set-variable frame-resize-pixelwise t)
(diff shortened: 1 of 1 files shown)


sthesing (approved):
Looks good to me. Thanks for handling this.
And interesting that here `featurep` actually works as intended :+1:

jeffbowman:
>  And interesting that here `featurep` actually works as intended 👍

Has to do with the fact `ns` is a built-in.  The `require` guard is needed when we need to pull in a library (built-in or not) so we can call a function that isn't autoloaded or to load a variable so we can bind a value to it.  Since `ns` gets loaded on a Mac by default, we don't need to `require` it.  Also, `locate-library` would not be useful since we know it is built-in and distributed with Emacs.  So, yeah, `featurep` for the win in this case! :smile:

Issue #346, opened 2023-09-10
[craftedv2beta] Evil bindings in crafted-speedbar are duplicate and inconsistent with evil-collection


jvdydev:
While writing the docs for `crafted-speedbar`, I noticed that the keybindings manually set for evil-mode (in the module) are also set by `evil-collection-speedbar-setup` which is called immediately after (if installed).

The keybindings in question:
| Function                       | Key(s) in crafted-speedbar | Key(s) in evil-collection |
|--------------------------------|----------------------------|---------------------------|
| speedbar-edit-line             | Return                     | Return                    |
| speedbar-toggle-line-expansion | Tab                        | Tab                       |
| speedbar-item-load             | C-Return                   | L                         |

I think the load on "L" is also a little less accident prone from loading (evaluating) random elisp.
The `crafted-evil` module also disables integration in favor of `evil-collection`, so should we just remove this entirely and let `evil-collection` handle it?

There are some other things (namely usage of `setq-default` for `defcustom` vars) I'll also change anyway while overhauling the module.

sthesing:
I use neither speedbar nor evil, so I cannot contribute to the question which keybinding makes more sense.

But regarding the logic of the module, I think it makes sense to remove evil-specific settings from the other modules and put them into `crafted-evil-config` behind a guard (e.g. `(with-eval-after-load "crafted-speedbar-config" ...`.

Users who use `evil` but don't want to use `crafted-evil-config` will want to figure out their bindings by themselves (without the other modules interfering). Users who want to use the `evil`-related settings of Crafted Emacs will load `crafted-evil`.

If we want to do this analogue to the way we did it with `crafted-completion`, maybe it's a sensible approach to put all the evil-specific things into `crafted-evil-config` but only load those redundancies with `evil-collection` if that package is not  installed?

```elisp
(if not (locate-library "evil-collection")
  ((with-eval-after-load 'crafted-speedbar-config
     ;;... evil speedbar settings
     )
   (with-eval-after-load 'crafted-completion-config
     ;;... evil vertico settings
     )
   ;; and so on...
   )
  (evil-collection-init))
```

Note: the evil vertico settings are just an example here, I haven't checked whether that overlaps with `evil-collection`.

jvdydev:
ye, I think it would also be a lot nicer to work on since anything that breaks with `evil` (no matter what other parts of Emacs are involved) is in the evil-config.
[...]

PR #348, opened 2023-09-10
[craftedv2beta] Bundle evil keybindings into `crafted-evil-config`

sthesing:
Load them only if `evil-collection` is not used.

### DIFF
--- a/modules/crafted-completion-config.el
+++ b/modules/crafted-completion-config.el
@@ -36,13 +36,6 @@ ARG is the thing being completed in the minibuffer."
   ;; Start Vertico
   (vertico-mode 1)
 
-  ;; configure keys for those who prefer vi keybindings
-  (when (featurep 'evil)
-    (with-eval-after-load 'evil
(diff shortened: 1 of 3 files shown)


sthesing:
As discussed in #346.
@jvdydev  feel free to ignore if this is more trouble then help :wink:

jvdydev:
Well, I already re-organized the speedbar module a bit.
Main problem is that for whatever reason there is a function called `speedbar-switch-to-quick-buffers` that really should be `crafted-speedbar-*` (cause it's a module function, not something that comes with speedbar) :laughing: 

So ye, appreciate the idea, I'll overhaul speedbar first, then work on the evil config separately (also obviously adds new stuff to the evil docs, potentially remove some stuff from completions docs, ...).

Thank you for the offer, hope you don't mind but I'll close this.

sthesing:
> Thank you for the offer, hope you don't mind but I'll close this.
Sure, that was the idea, after all.

PR #349, opened 2023-09-10
[craftedv2beta] Overhaul crafted-speedbar module

jvdydev:
General overhaul of the speedbar module to be more in-line with other modules in terms of coding style and documentation (speedbar module docs still WIP).

- Reorganize top-level into "Look & Feel", "Keybindings" and "File Extensions"
- Rename `speedbar-switch-to-quick-buffers` to `crafted-speedbar-switch-to-quick-buffers` to comply with naming convention
- Always map the switch-to-quick-buffers function to "b" (not just in evil-mode)
- Use `customize-set-variable` for `defcustom` variables instead of `setq-default`
- move speedbar-specific `evil-collection` initialization to `crafted-evil-config`
- update code comments
- Update documentation to reflect changes and regenerate info buffer

### DIFF
--- a/modules/crafted-evil-config.el
+++ b/modules/crafted-evil-config.el
@@ -83,11 +83,7 @@ Rebinds the arrow keys to display a message instead."
 
 (when (locate-library "evil-collection")
   ;; Initialize evil-collection
-  (evil-collection-init)
-
-  ;; If speedbar is loaded, run evil-collection setup
-  (with-eval-after-load 'speedbar
-    (evil-collection-speedbar-setup)))
+  (evil-collection-init))
 
 (provide 'crafted-evil-config)
 ;;; crafted-evil-config.el ends here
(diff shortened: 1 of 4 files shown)


modules/crafted-evil-config.el
```diff
@@ -81,7 +81,13 @@ Rebinds the arrow keys to display a message instead."
...
   (add-to-list 'evil-emacs-state-modes mode))
 
-(evil-collection-init)
+(when (locate-library "evil-collection")
+  ;; Initialize evil-collection
+  (evil-collection-init)
+
+  ;; If speedbar is loaded, run evil-collection setup
+  (with-eval-after-load 'speedbar
+    (evil-collection-speedbar-setup)))
```
sthesing, line 90: If I understood the docs of `evil-collection` correctly, then `evil-collection-init` already encompasses `evil-collection-speedbar-setup`. So that code can be removed.

sthesing (approved):
Looks good overall.
Only one bit of redundant code (see the other comment). Since that redundance doesn't have any repercussions, I approve this, regardless. Still probably cleaner to remove it.

Thanks!

PR #350, opened 2023-09-10
[craftedv2beta] Move evil-specific code to `crafted-evil`

sthesing:
See #346. Instead of #348.

- Move a few bindings from `crafted-completion-config`.
- Update documentation to reflect this.


jvdydev (approved):
Looks good to me.

PR #351, opened 2023-09-12
[craftedv2beta] Module documentation for `crafted-speedbar`

jvdydev:
Adds documentation for `crafted-speedbar` module.
Info buffer is already regenerated.

### DIFF
--- a/docs/crafted-speedbar.org
+++ b/docs/crafted-speedbar.org
@@ -51,11 +51,32 @@ It can be used for viewing file-trees, currently open buffers and more.
   Crafted Emacs adds a frame name/title, removes the minibuffer status
   and adds some border and fringe to the frame.
 
-- =new keybinding=: =speedbar-switch-to-quick-buffers= (=b=)
+  To reset the frame parameters:
+
+  #+begin_src emacs-lisp
+  (customize-set-variable 'speedbar-frame-parameters
+                          ((minibuffer)
+                           (width . 20)
+                           (border-width . 0)
+                           (menu-bar-lines . 0)
+                           (tool-bar-lines . 0)
+                           (unsplittable . t)
+                           (left-fringe . 0)))
+  #+end_src
+
+- =new keybinding=: =crafted-speedbar-switch-to-quick-buffers= (=b=)
 
   Temporarily change speedbar into a buffer switching tool,
   displaying the currently open buffers.
   Pressing =<enter>= on a buffer in the list will switch to that buffer.
+
+  To bind this function to another key:
+
+  #+begin_src emacs-lisp
+  (keymap-set speedbar-mode-map
+              "<key>"
+              #'crafted-speedbar-switch-to-quick-buffers)
+  #+end_src
 
 - =speedbar-add-supported-extension=: new extensions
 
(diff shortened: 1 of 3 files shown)


jeffbowman (approved):
Looks good to me.

docs/crafted-speedbar.org
```diff
@@ -0,0 +1,66 @@
...
+  #+end_src
+
+- =speedbar-frame-parameters=
+
+  Customizing =speedbar-frame-parameters= sets additional frame parameters
+  for speedbar frames.
+
+  Crafted Emacs adds a frame name/title, removes the minibuffer status
+  and adds some border and fringe to the frame.
+
```
sthesing, line 53: Maybe it would be a good idea to have an example of how the user could change this, e.g. back to the Emacs default values:
```elisp
(customize-set-variable 'speedbar-frame-parameters
  ((minibuffer)
   (width . 20)
   (border-width . 0)
   (menu-bar-lines . 0)
   (tool-bar-lines . 0)
   (unsplittable . t)
   (left-fringe . 0)))
```

docs/crafted-speedbar.org
```diff
@@ -0,0 +1,66 @@
...
+- =new keybinding=: =speedbar-switch-to-quick-buffers= (=b=)
```
sthesing, line 54: Shouldn't that be `crafted-speedbar-switch-to-quick-buffers`?

docs/crafted-speedbar.org
```diff
@@ -0,0 +1,66 @@
...
+
+  Temporarily change speedbar into a buffer switching tool,
+  displaying the currently open buffers.
+  Pressing =<enter>= on a buffer in the list will switch to that buffer.
+
```
sthesing, line 59: Maybe an example on how to bind another key to this function?

docs/crafted-speedbar.org
```diff
@@ -0,0 +1,66 @@
...
+- =speedbar-add-supported-extension=: new extensions
+
+  Crafted Emacs adds a wide selection of file extensions that support
+  tagging to speedbar's tagging system.
+  This allows speedbar to add tags when expanding a file entry with this
+  extension (e.g. ~org-mode~ headings) and make them clickable navigation
+  links.
```
sthesing, line 66: Here I'm not sure whether an example of how to change this is in order. The Emacs default is included in our settings and more doesn't hurt anybody. So it doesn't really make sense to put an example here. Aesthetically, I have the feeling that something's missing :shrug:

sthesing (changes requested):
This is really good, thanks.
I ask for an example for the user how to change this in a few places, and I think I spotted one error.
Otherwise this is fine by me. And then it's 14 down, one more to go :cartwheeling:

docs/crafted-speedbar.org
jvdydev: Don't really see what do add here, I'll resolve the rest, if you have any suggestions, I can add them

sthesing: You're right. I have no idea, either. 'Twas but a thought. Nevermind.

sthesing (approved):
Perfect. Thanks!

PR #352, opened 2023-09-13
[craftedv2beta] Add documentation for `crafted-init` customizations.

sthesing:
In #128, the getting started guide is named as the place for documentation for `crafted-init`. 
It is still lacking documentation for the two new customization variables `crafted-init-auto-save-customized` and crafted-init-auto-save-selected-passages`.

I couldn't think of a good heading, though. In the end, I wrote "To save or not to save customizations". Maybe someone has a better idea...?

### DIFF
--- a/docs/getting-started-guide.org
+++ b/docs/getting-started-guide.org
@@ -326,7 +326,7 @@ following lines to your config:
 
 #+begin_src emacs-lisp
   (customize-set-variable 'crafted-init-auto-save-customized nil)
-  (customize-set-variable 'crafted-init-auto-save-selected-passages nil)
+  (customize-set-variable 'crafted-init-auto-save-selected-packages nil)
 #+end_src
 
 * Where to go from here
--- a/docs/getting-started-guide.org
+++ b/docs/getting-started-guide.org
@@ -317,7 +317,7 @@ See the ~examples~ folder in the git-repo for more examples.
 
 * To save or not to save customizations
 
-As described above, by default Crafted Emacs will save both the list 
+As described previously, by default Crafted Emacs will save both the list
 =package-selected-packages= and all customizations to your customization file
 (e.g. ~custom.el~).
 
@@ -330,7 +330,7 @@ following lines to your config:
 #+end_src
 
 Whether or not to save these settings to ~custom.el~ is ultimately a matter of
-personal preference. You'll probably won't notice a difference either way.
+personal preference. You probably won't notice a difference either way.
 
 In the case of the customizations, we call the function
 =customize-save-customized= under the hood, which is originally intended for users
@@ -363,6 +363,10 @@ need to run those two functions.
 Still, if you prefer not to save customizations and/or the list of selected
 packages during each startup, you can turn that behaviour off as described
 above.
+
+Please note: These variables only affect whether /Crafted Emacs/ will or won't
+automatically store anything set by customize-set-variable, but that doesn't
+hinder other processes in Emacs to do so.
 
 * Where to go from here
 
--- a/docs/getting-started-guide.org
+++ b/docs/getting-started-guide.org
@@ -315,6 +315,20 @@ Example @@texinfo:@code{init.el}@@:
 
 See the ~examples~ folder in the git-repo for more examples.
 
+* To save or not to save customizations
+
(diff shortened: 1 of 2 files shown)


jvdydev:
Looks good, I actually like the heading.

I wonder if there would be any examples to add on when saving/not saving would be preferred?
(I remember something about a bug with hooks was discussed at some point)

docs/getting-started-guide.org
```diff
@@ -315,6 +315,20 @@ Example @@texinfo:@code{init.el}@@:
...
+As described above, by default Crafted Emacs will save both the list 
+=package-selected-packages= and all customizations to your customization file
+(e.g. ~custom.el~).
+
+You can customize that, too. To change that behaviour, add one or both of the
+following lines to your config:
+
+#+begin_src emacs-lisp
+  (customize-set-variable 'crafted-init-auto-save-customized nil)
+  (customize-set-variable 'crafted-init-auto-save-selected-passages nil)
```
jeffbowman, line 329:
```suggestion
  (customize-set-variable 'crafted-init-auto-save-selected-packages nil)
```

passages? lol

jeffbowman (changes requested):

sthesing:
> I wonder if there would be any examples to add on when saving/not saving would be preferred? (I remember something about a bug with hooks was discussed at some point)

The bug thing turned out to be unrelated, after all. @jeffbowman found out that it triggered regardless of that setting and then just showed different output. Still, good idea. I'll write a few lines. In the end, it comes down to personal preference, though.

docs/getting-started-guide.org
sthesing: Ugh! Good that you spotted that! Must've picked the wrong completion or something... Thanks!

sthesing:
> I wonder if there would be any examples to add on when saving/not saving would be preferred? (I remember something about a bug with hooks was discussed at some point)

Done. As mentioned, I left the bug out because it actually doesn't apply here, but tried to summarize our discussions about it.

docs/getting-started-guide.org
```diff
@@ -326,9 +326,44 @@ following lines to your config:
 
 #+begin_src emacs-lisp
   (customize-set-variable 'crafted-init-auto-save-customized nil)
-  (customize-set-variable 'crafted-init-auto-save-selected-passages nil)
+  (customize-set-variable 'crafted-init-auto-save-selected-packages nil)
 #+end_src
 
+Whether or not to save these settings to ~custom.el~ is ultimately a matter of
+personal preference. You'll probably won't notice a difference either way.
```
jeffbowman, line 333:
```suggestion
personal preference. You probably won't notice a difference either way.
```

Don't need the contraction in this case.

docs/getting-started-guide.org
```diff
@@ -315,6 +315,55 @@ Example @@texinfo:@code{init.el}@@:
 
 See the ~examples~ folder in the git-repo for more examples.
 
+* To save or not to save customizations
+
+As described above, by default Crafted Emacs will save both the list
```
jvdydev, line 320: I noticed that "As described above" doesn't really make sense in the info buffer (as there is nothing directly *above*.

Maybe "As described previously"?

jvdydev (commented):
Just something I noticed while reading the new version, not a major deal-breaker.

jeffbowman (changes requested):
There is another caveat which has nothing to do with Crafted Emacs.  Using `list-packagas` or if you have customizations in a `.dir-locals.el` file, Emacs will store updates to the `custom-set-variables` variable and write that to either `custom-file` if it is defined or `user-init-file` if it isn't.  Not setting the variable just means *Crafted Emacs* **won't automatically** store anything set by `customize-set-variable`, but there _are_ other processes in Emacs which will do so.

docs/getting-started-guide.org
sthesing: Good point. Thanks for spotting that. Fixed it.

docs/getting-started-guide.org
sthesing: Thanks! I overlooked that there are two "wills" in "you'll" and "won't". Fixed it.

jeffbowman (approved):

PR #353, opened 2023-09-17
[craftedv2beta] Use when instead of if, since else returns nil

jvdydev:
Small thing I found while checking through the code.

Using `if` with an else-branch that returns `nil` is equivalent to using `when`.
Arguably more readable.

```emacs-lisp
;; old
(if <test>
    ...
  nil)

;; new
(when <test>
  ...)
```


sthesing (approved):
Good catch

PR #354, opened 2023-09-17
[craftedv2beta] Prefer using format instead of concat

jvdydev:
We use format in most other contexts of similar nature. Especially with message, using the built-in formatting makes sense.

```emacs-lisp
;; old
(let ((mode "org"))
  (call-which-requires-mode-symbol (concat mode "-mode")))

;; new
(let ((mode "org"))
  (call-which-requires-mode-symbol (format "%s-mode" mode)))

;; old
(let ((mode "org"))
  (message (concat "I like " org "-mode")))

;; new
(let ((mode "org"))
  (message "I like %s-mode" mode))
```

I also added the formatting to print the path to custom-modules (like normal modules).

### DIFF
--- a/modules/crafted-ide-config.el
+++ b/modules/crafted-ide-config.el
@@ -36,8 +36,8 @@ manually with something like this:
                    (not (eq 'lisp-mode mode))     ; prefer sly/slime
                    (not (eq 'scheme-mode mode))   ; prefer geiser
                    )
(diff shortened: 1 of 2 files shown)


sthesing (approved):

jeffbowman (approved):
Nice! Much cleaner this way!

PR #355, opened 2023-09-17
[craftedv2beta] Use `#'` instead of `'` for passing functions

jvdydev:
We had some instances left over that used `'function` instead of `#'function`.
Fixing those.


jvdydev:
Fixed merge conflict, should be ready for merge now

Issue #359, opened 2023-09-17
texinfo tags in doc headings don't seem to be needed


sthesing:
In `CONTRIBUTING.org`, we currently write:

```
+ Use =@@texinfo:@code{}@@= or =@@texinfo:@samp{}@@= tags in
  headers, captions, or other places where org formatting does not
  directly translate correctly.
```

However, at least for org headings, all of the following are rendered identically into `` `foo` `` in the info file:
- `=foo=`
- `~foo~`
- `@@texinfo:@code{foo}@@`
- `@@texinfo:@samp{foo}@@`

Am I overlooking something here? If not, we can remove the "headers, captions or other" from the sentence above.
Removing the unnecessary texinfo tags should also mitigate #311 somewhat.

jvdydev:
I also noticed that (and from my testing, the doc export tool seems to also not care).
Might have been dependent on Emacs configuration (like some indentation things that we know change the exporter's functionality)?

In any case, I (personally) don't see a reason to keep them (at least for exporting to info).

jeffbowman:
Hmm.. They are there because when I first started writing docs, the places where I wanted to put `code` in headers didn't render the monospaced font as expected in the info files.  That was back in Emacs 27 and Rational Emacs days, so I haven't tried removing those markup tags in favor of staying with the `~` or `=` to make sure they rendered correctly.  Personally, I'd rather not have them in the org files since those are browsable on GitHub, but I *do* want the `code` rendering "correctly" in info files.  If that can happen with just the normal Org markers for `code` or `verbatim` that would be great and we can just remove the `@@texinfo...` tags.

sthesing:
closed by #360

PR #360, opened 2023-09-17
[craftedv2beta] Consistent org markup in docs

### LINKED ISSUE #311: texinfo tags in docs
Github doesn't process tags such as `@@texinfo:@emph{}@@`, so they just get displayed as part of the text. Kind of a minor issue, but many users will use Github to view the docs, especially when getting started.

sthesing:
use `=` for:

- file paths, e.g. `=init.el=`
- buffer names, e.g. `=*Messages*=`
- values (except numbers, which stay plain), e.g. `=nil=`
- keystrokes or keybindings, e.g. `=C-n=`
- text (to be) entered into the minibuffer (unless it is a complete command name), e.g. `=buf=`
- info nodes, e.g. `=(emacs)Just Spaces=`

`~` for:

- module names, e.g. `~crafted-defaults-config~`
- variable names, incl. external ones, e.g. `~load-prefer-newer~`, `~$PATH~`
- function names, e.g. `~car~`
- command names, incl. external ones, e.g. `~find-file~`, `~make~`
- package names, incl. ones that are shaped like a file name, e.g. `~denote~`, `~package.el~`

This resolves #311 #359

### DIFF
--- a/docs/crafted-osx.org
+++ b/docs/crafted-osx.org
@@ -21,11 +21,11 @@ to be set:
 
 | Keybinding | Meaning                          |
 |------------+----------------------------------|
-| ⌘-W       | (upper case W) delete-frame      |
-| ⌘-}       | tab-bar-switch-to-next-tab       |
-| ⌘-{       | tab-bar-switch-to-prev-tab       |
-| ⌘-t       | tab-bar-new-tab                  |
-| ⌘-w       | (lower case w) tab-bar-close-tab |
+| ⌘-W        | (upper case W) delete-frame      |
+| ⌘-}        | tab-bar-switch-to-next-tab       |
+| ⌘-{        | tab-bar-switch-to-prev-tab       |
+| ⌘-t        | tab-bar-new-tab                  |
+| ⌘-w        | (lower case w) tab-bar-close-tab |
 |------------+----------------------------------|
 
 We provide a function to set the frame titlebar transparent.  You call
--- a/docs/crafted-osx.org
+++ b/docs/crafted-osx.org
@@ -12,7 +12,7 @@ points.
 
 ** Description
 
-The =crafted-osx= module provides customizations to make Emacs more
+The ~crafted-osx~ module provides customizations to make Emacs more
 friendly on a Mac.  These largely have to do with keybindings as the
 Mac has a few keys different from a Linux or Windows computer.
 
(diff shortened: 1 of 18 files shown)


docs/crafted-osx.org
```diff
@@ -21,11 +21,11 @@ to be set:
 
 | Keybinding | Meaning                          |
 |------------+----------------------------------|
-| ⌘-W        | (upper case W) delete-frame      |
-| ⌘-}        | tab-bar-switch-to-next-tab       |
-| ⌘-{        | tab-bar-switch-to-prev-tab       |
-| ⌘-t        | tab-bar-new-tab                  |
-| ⌘-w        | (lower case w) tab-bar-close-tab |
+| ⌘-W       | (upper case W) delete-frame      |
```
jeffbowman, line 24: formatting looks off here.

docs/getting-started-guide.org
```diff
@@ -125,61 +122,58 @@ occur. Let's start by creating the ~init.el~ file in your
...
 
 ** Installing packages
 
-The standard approach to finding and installing packages is to use
-the command ~M-x list-packages RET~, which will bring up a
-user interface to search for packages, review the package details,
-install, update or remove package.
-For more information: [[info:emacs#Packages][Packages]] or on the web: [[https://www.gnu.org/software/emacs/manual/html_node/emacs/Packages.html][Packages]].
+The standard approach to finding and installing packages is to use the command
+=M-x= ~list-packages RET~, which will bring up a user interface to search for
```
jeffbowman, line 161:
```suggestion
~M-x list-packages RET~, which will bring up a user interface to search for
```

This one is awkward.  It renders in the info file a `M-x` `list-packages RET`, but it seems like it should render as `M-x list-packages RET`.  Maybe this should get the `~` treatment?  In the GitHub rendering, there is a space between the `M-x` and the `install-packages RET`.  I'd prefer there not be that space as this is a single command.

jeffbowman (changes requested):
A few changes to consider.

docs/getting-started-guide.org
jvdydev: I agree, I think personally this falls under both "keystrokes or keybindings" and "text (to be) entered into the minibuffer", so `=`, but I'm ok with `~` as well.

sthesing: I agree it can't stay like this. I'm also leaning towards `=`. And, come to think of it, we can get rid of the `RET` in this case, right?

sthesing: When I went through the docs to look for other occurrences of `=M-x= ~foo~`, I saw that we already had several occurrences of `=M-x foo=` already. I changed all others to match them, hope that's OK.
Thanks!

docs/crafted-osx.org
sthesing: Oh yeah, I see it now, too. The `⌘` causes the automatic table formatting in org to be a bit wonky, but I can see it when looking at the raw org in the github web interface. Thanks for pointing that out.

docs/getting-started-guide.org
jeffbowman: > I agree, I think personally this falls under both "keystrokes or keybindings" and "text (to be) entered into the minibuffer", so `=`, but I'm ok with `~` as well.

Ah... Yes, that works for me (agree with `=`).

jeffbowman (approved):
Approved. But dang it, I forgot I was supposed to merge this one first! I'll work out the conflicts in a few hours, unless someone beats me to it.

PR #364, opened 2023-09-22
Move creation of startup-screen to `crafted-startup-config`

sthesing:
I noticed that without this change, crafted-startup-config won't automatically create the splash screen, unless the user explicitely sets `initial-buffer-choice` to `#'crafted-startup-screen`. Which makes the variable `crafted-startup-inhibit-splash` redundant.

We only actually use it in the example init.el file.

With moving it to `crafted-startup-config`, it now behaves as documented. When the module is required, the splash screen is automatically shown, unless the user sets `crafted-startup-inhibit-splash` to non-nil.

### DIFF
--- a/examples/init-example.el
+++ b/examples/init-example.el
@@ -41,10 +41,6 @@
 ;; for you and add them here.
 (require 'crafted-defaults-config)
 (require 'crafted-startup-config)
-(unless crafted-startup-inhibit-splash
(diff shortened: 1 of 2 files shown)


jvdydev (approved):
Looks good to me, makes sense to align with the docs.

jeffbowman:
I can't remember why I didn't do that originally... so, good change to make! Thanks!

Issue #363, opened 2023-09-22
[v2] Emacs 29+ check for executable prevents Treesitter setup


jvdydev:
Emacs Versions affected: 29+
Tested on: Arch Linux, Emacs 29 custom w/ treesitter + native-comp

While tinkering with the RC1, I realized that treesitter is installed as a link library (at least on Arch), therefor the check for loading `treesit-auto` always returns nil for `(executable-find "tree-sitter")`.

How to reproduce (roughly):
- acquire Emacs 29 with treesitter enabled
- Configure `crafted-ide` module
- Run `M-: (featurep 'treesit-auto)`

Expected Result: `t`

Actual Result: `nil`

Proposal for fix: Remove the `(executable-find ...)` call
```emacs-lisp
;; before
(when (version< "29" emacs-version)
  (when (and (member "TREE_SITTER" (split-string system-configuration-features))
             (executable-find "tree-sitter"))
    ...))

;; after
(when (version< "29" emacs-version)
  (when (member "TREE_SITTER" (split-string system-configuration-features))
    ...))
```

sthesing:
> Proposal for fix: Remove the `(executable-find ...)` call

Hm, yeah, under the circumstances that seems to be the best option to make this work as intended.

jvdydev:
The problem is that I have no idea how this would behave for e.g. Emacs w/ treesitter in an environment where there is no treesitter library.
I also am not sure what platform/distribution actually has a "tree-sitter" executable.

While I did propose the change, I'm not sure it's the correct one to make here.
I may need to test more distros and see how it behaves with/without treesitter installed (and potential workarounds).

jvdydev:
Also `featurep` would reproduce this as the `(require 'treesit-auto)` never gets run (on a system with the treesitter lib), so the library is not loaded, hence `nil` is returned.

Once `(require 'treesit-auto)` is ran, the library is loaded, therefor `featurep` returns `t`.

`executable-find` searches for an executable in `$PATH`.

sthesing:
> The problem is that I have no idea how this would behave for e.g. Emacs w/ treesitter in an environment where there is no treesitter library. I also am not sure what platform/distribution actually has a "tree-sitter" executable.

I would guess that - just as Arch does it - sensible GNU/Linux distributions will list treesitter as a dependency for Emacs that has ts support. Windows might become an issue, though.

[...]

PR #365, opened 2023-09-22
[v2-RC1] Fix Treesitter guard in `crafted-ide` module

jvdydev:
Addresses #363.

Remove `(executable-find "tree-sitter")` call as it is distributed as a shared library.
Unwrap `(treesit-auto-install-all)` from `with-eval-after-load` as `treesit-auto` is loaded as part of the guard.

### DIFF
--- a/modules/crafted-ide-config.el
+++ b/modules/crafted-ide-config.el
@@ -67,14 +67,12 @@ Example: `(crafted-tree-sitter-load 'python)'"
 ;; Emacs versions after 29
 (when (version< "29" emacs-version)
   ;; only attempt to use tree-sitter when Emacs was built with it.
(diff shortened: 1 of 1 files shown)


sthesing (approved):
Looks good to me.

jeffbowman (approved):
Looks good!

jvdydev:
I'll double-check this works on Windows, then it can be merged :smile:

jvdydev:
Works on windows.

There is a caveat since it requires a C compiler installed (or more specifically, the `tree-sitter` Emacs package requires it be at least one of `'("cc" "gcc" "c99")`).
Same applies for Linux, of course and is not really related to this PR.

Issue #366, opened 2023-09-23
[v2] No easy way to opt out of tree sitter parsers


mgmarlow:
Hello! I'm on the RC1 branch on Windows 11 and Emacs 29.1. After configuring `crafted-ide` I ran into an issue where the `protobuf` parser failed to install (for some unknown reason). Since `crafted-ide-config` uses `treesit-auto-install-all`, every time I closed and reopened Emacs, I'd once again be prompted to install the missing tree sitter parser (protobuf) which would fail to install in the same way.

To work around this, I manually require `treesit-auto` and set the `treesit-auto-opt-out-list` before continuing with `crafted-ide-config`.

Perhaps Crafted Emacs should support its own method to avoid this install-loop and opt out of certain parsers?

```elisp
;; ...
(require 'crafted-ide-packages)

(package-install-selected-packages :noconfirm)

;; Require treesit-auto before loading ide-config so I can opt out of protobuf:
(require 'treesit-auto)
(add-to-list 'treesit-auto-opt-out-list 'protobuf)
(require 'crafted-ide-config)
```

jvdydev:
Thank you for trying out the RC1.

I understand that this can be quite annoying since it re-prompts on every boot if at least one grammar is not installed (it seems protobuf doesn't compile for me either :smiling_face_with_tear: ).

I think the best solution would be to make tree-sitter opt-in by default (e.g. with a `crafted-ide-configure-treesitter` function) which could take such a "opt-out" list of parsers.

Something like:
```emacs-lisp
(require 'crafted-ide-packages)

(package-install-selected-packages :noconfirm)

(require 'crafted-ide-config)
(crafted-ide-configure-treesitter '(protobuf)) ; configures treesitter, but without protobuf
```

This would also allow separating out the way to set up tree-sitter for Emacs pre-29 vs. 29-onwards.
Additionally it works around another limitation since tree-sitter requires a C compiler (which automatically makes Crafted Emacs have it as a requirement as well).

mgmarlow:
Agreed with the sentiment around requiring a C compiler, particularly for windows. I had previously set up tree sitter for myself before trying RC1 so I had WinGW ready to go. 

I would personally prefer an opt-in list as I really only care about tree sitter for a few languages, and it reads better from a configuration standpoint (the way the opt-out list is specified appears that protobuf is the only parser being installed, though the opposite is true). That's a minor nitpick though. 

[...]

PR #367, opened 2023-09-24
[v2] Make tree-sitter optional, add convenience functions

jvdydev:
Addresses #366.
Additionally addresses undocumented issue where tree-sitter requires a C compiler to compile grammars (somewhat hinted in #363).

Now, configuring tree-sitter works as follows:
```emacs-lisp

(require 'crafted-ide-config)

;; Emacs 28 or earlier
(crafted-ide-configure-tree-sitter)

;; Emacs 29 or later
;; To achieve the same behaviour as before
(crafted-ide-configure-tree-sitter)

;; Optionally pass a list of opt-out language grammars
(crafted-ide-configure-tree-sitter '(protobuf)) ; don't set up protobuf grammar
```

Documentation got two new sections (one for 28 and ealier, one for 29 and later).
Regenerated info buffer.

Note that this is a "fix" for these issues, although I'm not happy about the fact that language grammars are opt-in for pre-29 and opt-out for post-29. It should be either opt-in for both or neither.

### DIFF
--- a/docs/crafted-ide.org
+++ b/docs/crafted-ide.org
@@ -42,7 +42,46 @@ operating system. See [[https://github.com/joaotavora/eglot#connecting-to-a-serv
 
 Tree-Sitter support (built-in since Emacs 29) enables Emacs to better
 understand the syntax of your code, thus improving syntax highlighting and
-similar functions.
+similar functions. Usage of Tree-Sitter is optional.
+
+**** Configuring Tree-Sitter (Emacs 28 or earlier)
+
+Call ~crafted-ide-configure-tree-sitter~ after requiring ~crafted-ide-config~.
+
+#+begin_src emacs-lisp
+
+(require 'crafted-ide-config)
+(crafted-ide-configure-tree-sitter)
+
+#+end_src
+
+A new function ~crafted-tree-sitter-load~ will be available to
+install language grammars.
+
+#+begin_src emacs-lisp
+
+(crafted-tree-sitter-load 'python)
+
+#+end_src
+
+**** Configuring Tree-Sitter (Emacs 29 or later)
+
+Call ~crafted-ide-configure-tree-sitter~ after requiring ~crafted-ide-config~.
+This will install all known language grammars for Tree-Sitter.
+To opt-out of one or more language grammars, pass them as a list
+to ~crafted-ide-configure-tree-sitter~.
+
+#+begin_src emacs-lisp
+
+(require 'crafted-ide-config)
+
+;; install all language grammars
+(crafted-ide-configure-tree-sitter)
+
+;; install all language grammars, except protobuf
+(crafted-ide-configure-tree-sitter '(protobuf))
+
+#+end_src
 
 **** Combobulate
 
--- a/docs/crafted-ide.org
+++ b/docs/crafted-ide.org
@@ -44,6 +44,9 @@ Tree-Sitter support (built-in since Emacs 29) enables Emacs to better
 understand the syntax of your code, thus improving syntax highlighting and
 similar functions. Usage of Tree-Sitter is optional.
 
+To compile the language grammars on initial startup,
+a C compiler is required.
+
 **** Configuring Tree-Sitter (Emacs 28 or earlier)
 
 Call ~crafted-ide-configure-tree-sitter~ after requiring ~crafted-ide-config~.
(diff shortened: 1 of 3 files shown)


sthesing:
> Now, configuring tree-sitter works as follows:
> 
> ```emacs-lisp
> (require 'crafted-ide-packages)
> 
> (package-install-selected-packages :noconfirm)
> 
> (require 'crafted-ide-config)
> 
> ;; Emacs 28 or earlier
> (crafted-ide-configure-tree-sitter)
> 
> ;; Emacs 29 or later
> ;; To achieve the same behaviour as before
> (crafted-ide-configure-tree-sitter)
> 
> ;; Optionally pass a list of opt-out language grammars
> (crafted-ide-configure-tree-sitter '(protobuf)) ; don't set up protobuf grammar
> ```

I haven't had time to look at the code, yet. But just looking at the result,  #366 asked for an easy way to opt out of using tree-sitter. Now, it is opt-in for everyone. I think I'd prefer it to be an opt-out model.

> Note that this is a "fix" for these issues, although I'm not happy about the fact that language grammars are opt-in for pre-29 and opt-out for post-29. It should be either opt-in for both or neither.

Hm... I haven't really looked into the topic of language grammar installation, but it seems to me that the main problem to for making this opt-in or opt-put for everyone, regardless of the version they're using, is that there is no definite list of grammars that would need to be turned off or on in the respective other version. We could solve this by defining a list of "supported" language grammars. And that, I think, makes more sense for an opt-out model.

jvdydev:
ye, the reason I went for opt-in is it allows people to use `crafted-ide` on platforms without (easy) access to a C compiler (locked-down installations, Android, ...).
We had a similar discussion on `org-roam` a few months ago since it requires a locally compiled SQLite database (and we also don't auto-install `pdf-tools` for similar reasons).

I understand tree-sitter is *very nice* to have, but I don't think it's a good idea to enable it by default if it requires people to set up external things outside of Emacs (personally).

The only "opt-out" mechanism that would actually satisfy #366 would be a `defcustom` wrapping `treesit-auto-opt-out-list` (e.g. `crafted-ide-treesit-opt-out-list`):
```emacs-lisp

(customize-set-variable 'crafted-ide-treesit-opt-out-list '(protobuf))
(require 'crafted-ide-config)
```
This is because as shown in #366, `treesit-auto-opt-out-list` has to be modified in-between requiring `treesit-auto` and calling `treesit-auto-install-all`.

modules/crafted-ide-config.el
```diff
@@ -48,27 +48,32 @@ manually with something like this:
 (customize-set-variable 'eglot-autoshutdown t)
 
 ;;; tree-sitter
-;; Emacs versions prior to 29
-(when (version< emacs-version "29")
-  (when (require 'tree-sitter-indent nil :noerror)
+(defun crafted-ide--configure-tree-sitter-pre-29 ()
+  "Configure tree-sitter for Emacs 28 or earlier."
+  (when (version< emacs-version "29")
+    (when (require 'tree-sitter-indent nil :noerror)
```
jeffbowman, line 54: Why is this here?  The code which follows doesn't seem to need `tree-sitter-indent` unless I'm missing something - which is possible.

docs/crafted-ide.org
```diff
@@ -42,7 +42,46 @@ operating system. See [[https://github.com/joaotavora/eglot#connecting-to-a-serv
...
+
+(crafted-tree-sitter-load 'python)
+
+#+end_src
+
+**** Configuring Tree-Sitter (Emacs 29 or later)
+
+Call ~crafted-ide-configure-tree-sitter~ after requiring ~crafted-ide-config~.
+This will install all known language grammars for Tree-Sitter.
+To opt-out of one or more language grammars, pass them as a list
```
jeffbowman, line 71: Specifically mentioning the protobuf looping problem from issue #366 might be a useful example of when this might be applicable.

jeffbowman (changes requested):

modules/crafted-ide-config.el
jvdydev: Well, the version guard I just forgot to remove (as it's handled by `crafted-ide-configure-tree-sitter`).
As for the require guard, that was already there in the previous version.
I'll remove them both.

docs/crafted-ide.org
jvdydev: yes, I'll also add to the docs that using tree-sitter requires a C compiler to compile the language grammars.

jeffbowman (approved):

modules/crafted-ide-config.el
jeffbowman: D'oh!  Probably my fault then.  I wonder why I did that.  Thanks for fixing it!

PR #368, opened 2023-09-26
fix: version check for tree-sitter configuration

monora:
`crafted-ide--configure-tree-sitter-pre-29` should be called for versions < 29


jeffbowman (approved):
Woops!  Nice catch, thanks!

PR #370, opened 2023-09-27
Configure auto-mode-alist for tree-sitter modes

mgmarlow:
Add `auto-mode-alist` entries for many tree-sitter languages that have parsers but may lack the remap mode required by treesit-auto.

This is a bit of an experimental PR, but it solves a problem I ran into when setting up `crafted-ide`. I was surprised to learn that after installing every tree-sitter parser and using the `crafted-ide-config` functions for setting up tree-sitter, the `typescript-ts-mode` would not automatically activate when visiting `.ts` files. After some investigation, I found that in the `treesit-auto` [recipe definition](https://github.com/renzmann/treesit-auto/blob/main/treesit-auto.el#L243-L249) there's a `remap` field that corresponds to a specific major mode that `global-treesit-auto-mode` uses to map to the respective tree-sitter mode. Many of those remap modes are not installed by default.

For example, `typescript-ts-mode` only activates if you first install `typescript-mode` from MELPA. `global-treesit-auto-mode` looks for `typescript-mode` before it activates `typescript-ts-mode`. Otherwise, even with the tree-sitter parsers installed you will be in `fundamental-mode` when visiting `.ts` files.

I'm not 100% sure if this is something you all want for crafted-emacs, but I think it's a nice addition for those using tree-sitter for the first time. Visiting a `.ts` file should "just work" after the typescript parser is installed.

### DIFF
--- a/modules/crafted-ide-config.el
+++ b/modules/crafted-ide-config.el
@@ -75,11 +75,56 @@ OPT-OUT is a list of symbols of language grammars to opt out before auto-install
       ;; prefer tree-sitter modes
       (global-treesit-auto-mode)
       ;; install all the tree-sitter grammars
-      (treesit-auto-install-all))
+      (treesit-auto-install-all)
+      ;; configure `auto-mode-alist' for tree-sitter modes relying on
+      ;; `fundamental-mode'
+      (crafted-ide--configure-tree-sitter-modes))
     (when (locate-library "combobulate")
       ;; perhaps too gross of an application, but the *-ts-modes
       ;; eventually derive from this mode.
       (add-hook 'prog-mode-hook #'combobulate-mode))))
+
+(defvar crafted-ide-tree-sitter-auto-mode-alist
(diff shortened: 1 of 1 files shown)


jeffbowman:
well.. that's annoying.  I thought with tree sitter installed, then by default it would use `-ts-mode` instead of the just the normal language or (gasp) `fundamental` mode! 

Code looks good, my only concern is maintaining a variable mapping of modes (ie the `crafted-ide-tree-sitter-auto-mode-alist` value).  That is a minor concern.

mgmarlow:
On the topic of the map variable, I raised an issue in 'treesit-auto' as well (https://github.com/renzmann/treesit-auto/issues/55). My thought was if the file extensions are added to the 'treesit-auto' structs we could remove the map from crafted-ide entirely.

jeffbowman (approved):

jvdydev (approved):
Looks good to me :smile:

PR #372, opened 2023-09-29
docs: Add an evil-escape example

mgmarlow:
Add an example for binding alternative key combinations to escape from insert-mode when using crafted-evil.

I was hopeful this example would be useful for other Vim users who prefer rebinding Escape to "jj" or some other key combination.

It also addresses one confusion I had when working through the docs. When I add my own configuration outside of crafted-emacs, is it important that I avoid the use of something like `use-package` and prefer the `package-selected-packages` variable provided by crafted-emacs? For `evil-escape` at least, I ran into a few warnings/errors with my original `use-package` configuration that were resolved by swapping over to the `package-selected-packages` variable. The documentation is great for configuring the different aspects within crafted-emacs, but I was a little unsure how I should configure things outside of that zone.

### DIFF
--- a/examples/evil-escape.el
+++ b/examples/evil-escape.el
@@ -1,0 +1,45 @@
+;;; init.el -*- lexical-binding: t; -*-
+
+;;; Commentary:
+
+;; An example Emacs Init file that extends `crafted-evil-packages'
+;; with `evil-escape', allowing alternative key combinations to escape
+;; from Vim's insert-mode.
+
+(setq custom-file (expand-file-name "custom.el" user-emacs-directory))
+(when (and custom-file
+	   (file-exists-p custom-file))
+  (load custom-file nil :nomessage))
+
+(load "~/crafted-emacs/modules/crafted-init-config")
+
+;;; Packages phase
+
+(require 'crafted-evil-packages)
+
+;; Add `evil-escape', the package responsible for binding Escape to an
+;; alternative key combination.
+(add-to-list 'package-selected-packages 'evil-escape)
+
+(package-install-selected-packages :noconfirm)
+
+;;; Configuration phase
+
+(require 'crafted-evil-config)
+
+;; Configure `evil-escape' with the preferred key combination "jj".
+(customize-set-variable 'evil-escape-key-sequence (kbd "jj"))
+
+;; Allow typing "jj" literally without exiting from insert-mode
+;; if the keys are pressed 0.2s apart.
+(customize-set-variable 'evil-escape-delay 0.2)
+
+;; Prevent "jj" from escaping any mode other than insert-mode.
+(defun my/not-insert-state-p ()
+  "Inverse of `evil-insert-state-p`"
+  (not (evil-insert-state-p)))
+
+(customize-set-variable 'evil-escape-inhibit-functions (list #'my/not-insert-state-p))
+
+;; Enable `evil-escape' mode globally.
+(evil-escape-mode)


jeffbowman:
> is it important that I avoid the use of something like use-package and prefer the package-selected-packages variable provided by crafted-emacs?

Yes and no, or "it depends".  You may use `use-package` if you wish, however, where you do that makes a difference from a Crafted Emacs perspective.   Crafted Emacs considers these 3 "phases" in this order: packages, config, user.   Thus if you choose to have `(require 'crafted-evil-packages)` in your configuration, the expectation is you will have a `(package-install-selected-packages)` somewhere later in your configuration, but before _any_ configuration.  The idea is that you need to have the packages installed before you can configure them, with eg. `(require 'crafted-evil-config)`.  So, if you wish to use `use-package`, you might consider to provide your own custom module (conveniently located in the `user-emacs-directory / custom-modules` folder), where you handle all your evil package installation and configuration.  In this case, you would have the normal package section at the top, the configuration section in the middle, then finally the user configuration where you use your own modules/methods for configuring anything left over or not provided by Crafted Emacs.  

Thus (for example - note that you aren't installing `crafted-evil-packages`):

    (require 'crafted-ide-config)
    (use-package 'evil
      :demand t
      :config (...))

Or possibly:

    (require 'crafted-ide-config)
    (use-package 'evil :demand t)
    (use-package 'evil-collection :demand t :after 'evil)
    (use-package 'evil-nerd-commenter :demand t :after 'evil)
    (require 'crafted-evil-config)  ;; or put this in one of the above :config blocks - up to you.

These are just ideas how you might choose to use `use-package` in conjunction with Crafted Emacs.

examples/evil-escape.el
```diff
@@ -0,0 +1,39 @@
...
+(require 'crafted-evil-config)
+
+;; Configure `evil-escape' with the preferred key combination "jj".
+(setq evil-escape-key-sequence "jj"
+      ;; Allow typing "jj" literally without exiting from insert-mode
+      ;; if the keys are pressed 0.2s apart.
+      evil-escape-delay 0.2
+      ;; Prevent "jj" from escaping any mode other than insert-mode.
+      evil-escape-inhibit-functions (list
+				     (lambda () (not (evil-insert-state-p)))))
```
jvdydev, line 37: Not a request for change, more a point of discussion.

We (try to) follow a general [code style for Emacs](https://github.com/bbatsov/emacs-lisp-style-guide#functions) (see also #280).
Given that, this should probably be something like:
```emacs-lisp
(defun my/not-evil-insert-state-p ()
  "Inverse of `evil-insert-state-p`"
  (not (evil-insert-state-p)))

(customize-set-variable 'evil-escape-inhibit-functions (list #'my/not-evil-insert-state-p))
```

@jeffbowman thoughts?

jvdydev (changes requested):

examples/evil-escape.el
jeffbowman: Agree.  

But the name should probably be `crafted-evil-not-insert-state-p`  rather than `my/...`.

jvdydev: Makes sense, I used `my/...` as it's an example (therefor doesn't belong in the `crafted-evil` module), but also makes sense to not use `my/` for stuff we wrote.

jeffbowman: Oh... I thought the point was to have that in `crafted-evil`!  Yeah, if it is supposed to be in the user's config, `my/...` is perfectly fine.

examples/evil-escape.el
```diff
@@ -0,0 +1,39 @@
...
+(add-to-list 'package-selected-packages 'evil-escape)
+
+(package-install-selected-packages :noconfirm)
+
+;;; Configuration phase
+
+(require 'crafted-evil-config)
+
+;; Configure `evil-escape' with the preferred key combination "jj".
+(setq evil-escape-key-sequence "jj"
```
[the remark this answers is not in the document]
mgmarlow: I think originally I was going [off of the docs](https://github.com/syl20bnr/evil-escape#key-sequence), but your `(kbd "jj")` works as well, updated 👍

examples/evil-escape.el
mgmarlow: Yeah the intention is for this to be user-land, I'd imagine it's a frequent change but better as an example than something crafted-emacs would handle imo

jvdydev (approved):
Just saw this, looks good to me now.

jeffbowman (approved):

PR #373, opened 2023-09-29
Splash screen updates

jeffbowman:
Some updates to add "modules" like the list of projects (from `project.el`) and today's diary entries to the splash screen.  Need to customize the list `crafted-startup-module-list` to call the functions which put the information on the screen.  Needs to be added to documentation, this PR is for reviewing the code.

### DIFF
--- a/modules/crafted-startup-config.el
+++ b/modules/crafted-startup-config.el
@@ -122,9 +122,6 @@ Each element in the list should be a list of strings or pairs
   (when (file-exists-p project-list-file)
     (project--read-project-list)
     (message "Showing projects on splash screen")
-    (message (format "project count %s" (length project--list)))
-    (dolist (proj (seq-take project--list crafted-startup-project-count))
-      (message (format "%s" (car proj))))
     (fancy-splash-insert
      :face '(variable-pitch font-lock-string-face italic)
      (condition-case project--list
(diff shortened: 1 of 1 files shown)


modules/crafted-startup-config.el
```diff
@@ -88,30 +102,76 @@ Each element in the list should be a list of strings or pairs
...
+                   (overlay-put button 'checked t)
+                   (overlay-put button 'display
+                                (overlay-get button :on-glyph))
+                   (setq crafted-startup-screen-inhibit-startup-screen t))))))
   (fancy-splash-insert :face '(variable-pitch (:height 0.9))
                        " Never show it again."))
 
+(defun crafted-startup-diary ()
+  (require 'diary-lib nil :noerror)
+  (when (diary-check-diary-file)
```
jvdydev, line 131: Not sure it's intended, but this errors when there is no diary file.
This results in the entire startup screen not loading.

Given it's used as a guard here, I think `(file-readable-p diary-file)` makes more sense?

jvdydev (changes requested):
Just one thing I noticed (diary startup module crashes when there is no diary file).

modules/crafted-startup-config.el
jeffbowman: Shucks... missed that one.  Yes, the `diary-check-diary-file` throws errors, I was just trying not to re-implement that function, but I guess I need to.  Thanks for point it out.

jvdydev (approved):

PR #376, opened 2023-10-01
[v2] (crafted-ide) Make aggressive-indent-mode optional

mgmarlow:
After working with crafted-emacs for a week or so, I've found `aggressive-indent-mode` to be, well, too aggressive. It works great for all Lisp modes (hence crafted-lisp adding it there as well) but for general programming I've found that it causes more issue than it solves.

The common problematic case for me is editing code where style conventions are set by an external library (think Prettier or Rubocop), or may not be set at all. `aggressive-indent-mode` will change indentation across the whole file when making any kind of edit, messing up a previous contributor's indentation to fit my Emacs indentation preferences. I then have to disable `aggressive-indent-mode` in the buffer, make my edits, etc. Electric works better in these situations because it's localized to my changes, not the entire file.

I kept the defcustom `t` by default to match the existing preference and make it opt-out.

Maybe out of the scope of this PR but a good topic for discussion, this change disables a package that is installed via `crafted-ide-packages`, effectively negating it. In most cases I'd probably prefer to not install such a package in the first place. Is there somewhere else to put this option so the configuration can be shared and known across both the packages & config files? Is that desired?

### DIFF
--- a/docs/crafted-emacs.info
+++ b/docs/crafted-emacs.info
@@ -2062,20 +2062,28 @@ The ‘crafted-ide’ module sets up (and installs if necessary)
 functionality that makes Emacs act like a integrated development
 environment (IDE).
 
-   It sets up project based buffer management (ibuffer-project
-(https://github.com/muffinmad/emacs-ibuffer-project)) and a
-semi-automatic indentation (aggressive-indent
-(https://github.com/Malabarba/aggressive-indent-mode)) for many
-programming and markup languages.
+   This module includes:
 
-   If you use EditorConfig (https://editorconfig.org), this module
-provides support for that, too (see editorconfig-emacs
-(https://github.com/editorconfig/editorconfig-emacs) for details).
+   • Project-based buffer management (ibuffer-project
+     (https://github.com/muffinmad/emacs-ibuffer-project))
+   • LSP client (Eglot)
+   • Next-generation syntax parsing (Tree-Sitter)
+   • EditorConfig (https://editorconfig.org) support (editorconfig-emacs
+     (https://github.com/editorconfig/editorconfig-emacs)).
 
-   Furthermore, it sets up Eglot (a LSP-client) and Tree-Sitter (a next
-generation syntax parser).
+  1. Aggressive Indentation
 
-  1. Eglot
+     By default Emacs enables Electric Indent
+     (https://www.gnu.org/software/emacs/manual/html_node/emacs/Indent-Convenience.html),
+     a minor mode that automatically indents code after a RET key is
+     pressed.  ‘crafted-ide’ adds support for Aggressive Indent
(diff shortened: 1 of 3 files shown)


jvdydev:
I like the idea, however I'm not sure that it really requires a `defcustom`.
With v2, we try to stay away from too much crafted-emacs-specific Variables to make it easier to move out parts of the configuration into your own.

I'd prefer leaving the `aggressive-indent` configuration in-place and maybe swapping it to be opt-in.
This way, it can be enabled for all prog-modes by the user with a simple `add-hook` (which would have to be documented).

Regarding the general PR, I also want to point out that if the user opts out of `aggressive-indent` in this implementation, it will not only not be enabled, but also not be loaded.
I'm not sure this would be the behavior we'd want.

Anyway, just my thoughts for discussion.

mgmarlow:
That all sounds good to me, agreed on the defcustom. By "leaving the configuration in-place" are you referring to the `require` statement? As far as I understand a user who has `crafted-ide-packages` would only need to register the hook since `aggressive-indent-mode` is autoloaded. I'm imagining that the `aggressive-indent` require statement could be removed from `crafted-ide-config` altogether, with the addition of user instruction to enable it via a hook if desired.

jeffbowman:
Hmm... good points.  My development has been largely relegated to lisps and some python.  Which, depending on the client I'm working for, may introduce third party code conventions ala `blacken`.  That is usually handled in a `Makefile`, but Emacs has a `blacken-mode` and it does tend to argue with `aggressive-indent` on occasion.  Other languages which might be problematic would those like `go` which have their own ecosystem of tooling around formatting code.  Maybe we are too aggressive in providing this configuration for **all** programming modes.  I can't say I've run into too many problems with it enabled, but I can certainly understand places where it can get in the way.   

Maybe we leave it in the `crafted-ide-packages` module, but then provide documentation on adding it to a mode hook rather than providing the `prog-mode` hook as we do now.  @jvdydev, @sthesing thoughts on this?

sthesing:
I agree with the points raised. We're talking about `crafted-ide-config`, though right? Not `crafted-ide-packages`.

jeffbowman:
> I agree with the points raised. We're talking about `crafted-ide-config`, though right? Not `crafted-ide-packages`.

Yes, because `crafted-ide-packages` will not change, but instead we change `crafted-ide-config` module and appropriate documentation.

mgmarlow:
The new changes I posted remove `aggressive-indent-mode` from the `crafted-ide-config`. I also removed the configuration for Electric Indent since I read that it's been enabled by default since Emacs 24.4, but lmk if you'd like me to keep that in place.

I tweaked some of the surrounding structure for the crafted-ide documentation since the lack of aggressive-indent-mode made it a tad awkward, hope that's OK.

jeffbowman (approved):
Looks good to me.  Thanks!

PR #380, opened 2023-10-02
[v2] (crafted-completion) Use default corfu-auto-delay

mgmarlow:
Closes https://github.com/SystemCrafters/crafted-emacs/issues/369. Remove `crafted-completion`'s variable customizations for `corfu-auto-delay` and `corfu-echo-documentation` in favor of documentation.

BTW `corfu-echo-documentation` appears to be dropped in more recent corfu releases, see https://github.com/minad/corfu/blob/main/CHANGELOG.org#version-029-2022-11-19. It's now called `corfu-echo-delay`.


jeffbowman (approved):

sthesing:
Thanks, @mgmarlow !

PR #386, opened 2023-10-09
Prefer treesit-auto library function for configuring auto-mode-alist

mgmarlow:
https://github.com/renzmann/treesit-auto/pull/56 adds a new function, `treesit-auto-add-to-auto-mode-alist`, that automatically adds any ready treesitter grammars to the `auto-mode-alist` variable. It's now available in the latest release (0.6.5) so it can replace the manual implementation in crafted-emacs.


jeffbowman (approved):

jeffbowman:
Nice!  I appreciate you working with the `treesit-auto` package maintainer to get a general solution more people can use!  Thanks for the update!

PR #396, opened 2023-11-30
Add crafted-ide-eglot-auto-ensure-all for enabling eglot-ensure

mgmarlow:
Closes https://github.com/SystemCrafters/crafted-emacs/issues/395. After loading Eglot successfully, register `eglot-ensure` hooks for all relevant LSP modes.


jeffbowman:
This meets the description of your PR, I would like to consider what a solution would look like which allows a user to use the crafted-ide module _only_ for the tree-sitter configuration.  Currently, that works out as a side-effect of never loading `eglot`.  However, it might be nice to continue to allow that behavior and still solve the issue of the hooks only being loaded after `eglot` loads.  Perhaps wrap in a `defun` and allow the user to call it?

jeffbowman:
Good stuff, I like it!  The only thing I would request on this PR would an update to the docs, see `docs/CONTRIBUTING.org` for information on doing that. 

Thanks!

docs/crafted-ide.org
```diff
@@ -41,10 +41,25 @@ typed. You can enable it for programming modes with the following:
...
+installed through your operating system. See [[https://github.com/joaotavora/eglot#connecting-to-a-server][this list]] of servers that
+work with Eglot out of the box.
+
+**** Automatically start Eglot when visiting language buffers
+
+It's convenient to add a call to ~eglot-ensure~ for major modes that
+have an accompanying language server so that Eglot automatically
+initializes when that major mode is activated. ~crafted-ide~ makes a
+function available that will automatically register such ~eglot-ensure~
+hooks for all major modes with a detected language server. You can
```
jeffbowman, line 56:
```suggestion
function available that will automatically add ~eglot-ensure~
to all major mode hooks for which a language server is installed and detected. You can
```

jeffbowman (changes requested):
Minor wording change suggested.

jeffbowman (approved):

jeffbowman:
Thanks very much for your contribution!

PR #398, opened 2023-12-03
change git clone instruction

bird-dancer:
clone from master as it is the new default for v2
perform a shallow clone to save bandwidth and disk space


jeffbowman:
Oops!! I missed that one! Thanks for catching it!

PR #401, opened 2023-12-10
Add crafted-contrib module

mgmarlow:
Getting the ball rolling on https://github.com/SystemCrafters/crafted-emacs/discussions/399. Add a `crafted-contrib` module that adds a couple of useful packages, namely `debbugs`, `package-lint` (& `package-lint-flymake`), and `aggressive-indent`. My goal with this module is to help folks get started contributing to packages/Emacs core. For me personally, most of the value comes from enabling Flymake w/ `package-lint`, since that covers most of the Emacs Lisp basic formatting/code style/commenting guidelines that are necessary for creating and maintaining Emacs Lisp packages.

I also brought over `aggressive-indent` because I don't personally use `crafted-lisp`--I generally only write Emacs Lisp (and Fennel/Janet) so there's no need for me to install Common Lisp/Clojure/Scheme packages. I think `aggressive-indent` still fits within the "code style" goal of `crafted-contrib`.

I'm very interested in hearing if there are any other config/packages that you all think help with Emacs Lisp contribution, and what plans are for the crafted-emacs-specific tooling. As it stands this module is pretty light on features.


jeffbowman:
Sorry for late response to this.  I added some thoughts to #399 which would apply here.

mgmarlow:
Closed by https://github.com/SystemCrafters/crafted-emacs/pull/404

PR #403, opened 2023-12-20
Change symantics for treesit-auto configuration to opt-in

### LINKED ISSUE #402: treesit-auto-opt-out-list seems to have disappeared
Now getting this warning when starting Emacs:

```
configuring treesit-auto
 ■  Warning (initialization): An error occurred while loading ‘/home/jeffbowman/.emacs.d/init.el’:

Symbol's value as variable is void: treesit-auto-opt-out-list

To ensure normal operation, you should investigate and remove the
cause of the error in your initialization file.  Start Emacs with
the ‘--debug-init’ option to view a complete error backtrace.
```

Versions up to 0.6.6 don't have this problem, while upgrading treesit-auto to 0.6.7 removes the variable.  To limit the list of languages for which treesit-auto installs grammars, customize the `treesit-auto-langs` value to just the languages desired.  The documentation gives the example ` '(python go rust) ` to auto install those grammars when buffers of those modes are visited.

jeffbowman:
Versions of treesit-auto after 0.6.6 remove the variable treesit-auto-opt-out-list which we were using to opt out of certain languages that caused problems, for example proto/protobuf.

treesit-auto always seemed to provide a way to limit the installed ts modes to just the list in treesit-auto-langs, thus the user should customize this variable appropriately, or accept the default of all language ts modes.

crafted-ide--configure-tree-sitter changed to use the optional provided list as the "opt-in" list rather than the "opt out" list.

Fixes #402

### DIFF
--- a/modules/crafted-ide-config.el
+++ b/modules/crafted-ide-config.el
@@ -82,15 +82,21 @@ Example: `(crafted-tree-sitter-load 'python)'"
            (intern (format "%s-mode-hook" (symbol-name lang-symbol)))))
       (add-hook mode-hook-name #'tree-sitter-mode))))
 
-(defun crafted-ide--configure-tree-sitter (opt-out)
+(defun crafted-ide--configure-tree-sitter (opt-in-only)
   "Configure tree-sitter for Emacs 29 or later.
-OPT-OUT is a list of symbols of language grammars to opt out before auto-install."
+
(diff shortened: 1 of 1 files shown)


sthesing (approved):
Looks good to me. Thanks!

jvdydev (approved):

PR #405, opened 2023-12-28
Add tips and tricks docs for easier module loading

mgmarlow:
When there was some discussion around https://github.com/SystemCrafters/crafted-emacs/issues/321 I had meant to open a PR for an easier way to load modules, but didn't get around to it until now. IMO I'd probably lean towards including this function in crafted-emacs proper, since the configuration is otherwise a little unwieldly and it's easier to introduce backwards-incompatible changes with a top-level function that can be tweaked according to the change. But including it in the docs is a nice intermediate solution.

I'm not 100% sure the spot I picked for "Tips and tricks" is the right one, so happy to move it around if ya'll think it's better someplace else.


jeffbowman:
Tips and tricks section is fine.  I'd probably drop the `post install` bits as there are many modules with optional post install configuration (writing, evil, startup, etc). 

Alternatively, if that portion stays, then probably expand the comments above it to indicate this is an example and that other modules might have post-install configuration functions available.

mgmarlow:
> I'd probably drop the `post install` bits as there are many modules with optional post install configuration (writing, evil, startup, etc).

Yeah I was curious your thoughts on that section. I thought it was neat for my own purposes but it's probably not worth keeping it around since it's yet another spot to update when things change.

sthesing:
Hey @mgmarlow  thanks for the idea! I understand where you're coming from, I'm doing something similar in my config (albeit with a different approach).

For the docs, I usually take the perspective of (new) Emacs users without much elisp literacy and/or technical expertise in general (because that's what I am :grin: ). From that perspective, this example would need a bit more explanation, e.g.:
- What problem does this helper function try to solve?
- What are the use cases in which I can just use this as is?
- What do I need to do if I just want the packages but not the config modules? Or vice versa?
- What happens if I call this without a `modules` argument? Can I just redefine the `defaults` or will this break things?
- How do I apply this to my config?

I'd be happy to be of any help here (give feedback or even co-write this with you, if you want).

mgmarlow:
Good tips @sthesing, I updated the docs accordingly but feel free to suggest or make any other edits that you find helpful. I also dropped the optional argument because it seemed a bit unnecessary for this section--were this function provided by Crafted Emacs itself I could see the usefulness of default modules, otherwise it's just extra noise.

sthesing:
Sorry, @mgmarlow, I totally lost track of this.
Good changes, this is much better. I might have a few ideas to improve further upon them, but I feel it's best to merge this now, because it is already very helpful to users as is.
Would you mind to regenerate the info file to resolve conflicts with the merges that have happened in the meantime? Then I'll merge it.
Again, sorry for the delay and thanks for the contribution.

mgmarlow:
> Sorry, @mgmarlow, I totally lost track of this.

No problem! Rebased off main so everything should be good to go.

PR #406, opened 2024-01-04
Fix error passing nil to load

jvalleroy:
To avoid passing "nil" to the load function, use when instead of unless.

This change fixes the error for me, so I hope it is correct!


jeffbowman (approved):
Wow what a goof! Even the comment indicates a `when` should be used.  Thanks for finding this one!

PR #407, opened 2024-01-11
crafted-ui: Add breadcrumbs

mgmarlow:
[breadcrumb](https://github.com/joaotavora/breadcrumb) is an excellent little package that adds file descriptors to the top of the window (header line). IMO breadcrumbs are something that Emacs is sorely lacking and this package synergizes incredibly well with Eglot, though Eglot is not a requirement (https://github.com/joaotavora/eglot/discussions/988).

Here's what it might look like w/ a split-window view of a Rust and ELisp buffer.

breadcrumbs


jeffbowman:
Looks good to me... I don't typically use breadcrumbs because I have the modeline for following the function I'm in... not a bad addition though... I let @sthesing comment

PR #410, opened 2024-02-21
Don't crash when elisp-demos is not available

Bost:


jeffbowman (approved):

jeffbowman:
Looks like we missed one, thanks for catching it!

PR #413, opened 2024-04-13
Fix broken startup screen, reintroduce `concise`

sthesing:
449bde3 breaks the startup screen in `crafted-startup-config` by removing the optional parameter `concise`.
This reintroduces it.

### DIFF
--- a/modules/crafted-startup-config.el
+++ b/modules/crafted-startup-config.el
@@ -156,7 +156,7 @@ splash screen as a logo."
       (fancy-splash-insert :face logo-face (concat spaces line "\n")))
     (insert "\n")))
 
-(defun crafted-startup-tail ()
+(defun crafted-startup-tail (&optional concise)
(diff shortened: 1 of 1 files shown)


jeffbowman:
Well, dang it... I couldn't find it being used and so I removed it... my splash screen still worked... but maybe I had a cached version.   Thanks for fixing my blunder!

sthesing:
So did mine, but I am always the guardian of working examples. And our example init calls it with the concise option somewhere, so I noticed.
Back then I left it in because it is a option that the vanilla splash screen has. We _could_ get rid of it, but I figured it might be easier for users to adapt custom setups to the vanilla splash to our splash...

jeffbowman:
Yeah, that's a good point.  I'll add a comment about keeping the `&optional; concise`  parameter so I don't think I can remove it again later... still :facepalm:

Issue #415, opened 2024-05-29
`(require 'crafted-osx-config)` does nothing


jf:
### Operating System

Mac

### Emacs Version

29.1

### Emacs Configuration Details

- [X] Native Compilation
- [ ] pGTK
- [ ] alternative package manager (use-package, straight.el, ...)

### Anything else that may be related to the issue you're having?

maybe this matters... and maybe it doesn't (I'm guessing it shouldnt really), but I'm using emacs-mac-app 10.0 (https://ports.macports.org/port/emacs-mac-app/), installed via macports

### What happened?

I've got crafted-emacs set up as per https://github.com/SystemCrafters/crafted-emacs/blob/master/docs/getting-started-guide.org, and seeing as how there is a `crafted-osx-config.el`, I go ahead to do `(require 'crafted-osx-config)`.

This seems to achieve nothing, though. Reading the code, I expect `⌘-t` to give me a new tab... instead it transposes words. I dont see a `crafted-osx-packages.el`... so I'm not expecting to have to install any packages to support the c onfig.

Minimal demo `init.el`:
```
(setq custom-file (expand-file-name "custom.el" user-emacs-directory))
(when (and custom-file
	   (file-exists-p custom-file))
  (load custom-file nil))

(load "~/.config/crafted-emacs/modules/crafted-init-config")

(require 'crafted-osx-config)
```

### What should have happened?

see "What happened?"

sthesing:
@jf Thanks for opening this. `crafted-osx-config` is one of the things we can't test very well ourselves.

The way we check whether a user is on MacOS is by checking for the feature `ns` (which stands for NextStep, I believe). An explanation why the module doesn't do anything might be that `(featurep 'ns)` evaluates to `nil` in your setup.
Can you check that for us? Maybe we just need to change the conditions in the code.

In the meantime, as a workaround, you can try to copy the contents of the module to your config and remove the `(when (featurep 'ns)` and the corresponding `)` after the wrapped code and see whether it works as intended.

jf:
ok:

1. it looks like your macos detection is one fault here: evaluating `(featurep 'ns)` in a scratch buffer in Emacs gives me `nil`.
2. who wrote or put out this code then? this code really should have been tested.
3. no ELisp expert here, but a search shows that I can check `system-type`, and at least on my system, it works: `(eq system-type 'darwin)`

sthesing:
[...]

PR #416, opened 2024-05-31
Fix crafted-osx-config

jf:
attempts a fix at #415

### DIFF
--- a/modules/crafted-osx-config.el
+++ b/modules/crafted-osx-config.el
@@ -22,10 +22,10 @@
 
 ;;; Special keys
 (when (eq system-type 'darwin)
-  (customize-set-variable mac-right-option-modifier nil)
-  (customize-set-variable mac-command-modifier 'super)
+  (customize-set-variable 'mac-right-option-modifier nil)
+  (customize-set-variable 'mac-command-modifier 'super)
   (when (featurep 'ns)
-    (customize-set-variable ns-function-modifier 'hyper))
+    (customize-set-variable 'ns-function-modifier 'hyper)))
 
 ;;; Keybinds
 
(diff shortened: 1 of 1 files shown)


jeffbowman:
I looked at this, it seems reasonable, I'd like @jvdydev and/or @sthesing to also review.

jvdydev:
Looks good!

(just a general thought)
I wonder if we can drop the `(interactive)` on `crafted-osx-transparent-titlebar` as it's probably more of a configuration-time change? (also there is no way to reverse it)

jvdydev (approved):

jf:
> (just a general thought) I wonder if we can drop the `(interactive)` on `crafted-osx-transparent-titlebar` as it's probably more of a configuration-time change? (also there is no way to reverse it)

thank you for bringing that up! I've been wondering if that's even needed myself. From the help for `interactive`, Emacs says `Specify a way of parsing arguments for interactive use of a function.`. Sorry, newbie question, but what would interactive use of `crafted-osx-transparent-titlebar` look like? Would anybody use it that way?

sthesing (approved):
Looks good to me.
Thanks for preparing this.

jvdydev:
> > (just a general thought) I wonder if we can drop the `(interactive)` on `crafted-osx-transparent-titlebar` as it's probably more of a configuration-time change? (also there is no way to reverse it)
> 
> thank you for bringing that up! I've been wondering if that's even needed myself. From the help for `interactive`, Emacs says `Specify a way of parsing arguments for interactive use of a function.`. Sorry, newbie question, but what would interactive use of `crafted-osx-transparent-titlebar` look like? Would anybody use it that way?

Oh, sorry, forgot to explain. Marking a function `(interactive)` essentially means it's executable using `M-x the-function`. :smile:

PR #422, opened 2024-06-09
Streamline customization subgroups

sthesing:
When answering #420, both @jvdydev and I came to the conclusion that some of our uses of `defcustom` is inconsistent. I started out to fix this by removing a few implicitly defined customization groups like `crafted-faces` and streamlined them into a structure where all Crafted Emacs' customizations are organized in subgroups of `crafted` that adhere to the same naming scheme as the module in which they are defined. So a defcustom use in `crafted-defaults-config.el` is part of the subgroup `crafted-defaults`. I believe that makes the most sense. You folks think so, too, @jeffbowman and @jvdydev ?

I also added respective documentation in "Contributing".

### DIFF
--- a/docs/crafted-emacs.info
+++ b/docs/crafted-emacs.info
@@ -1087,7 +1087,23 @@ your idea is in the spirit of the *note Principles::.
      values.  This helps make sure constructors or setters attached to
      the variable are run when the value is set.
    • Provide ‘defcustom’ variables for things we expect the user to
-     modify and make sure it is in the appropriate group.
+     modify.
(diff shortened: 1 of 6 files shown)


jeffbowman (approved):
These changes look good to me.

jeffbowman:
I actually thought we were already doing this, but I guess we missed some stuff when we created v2. :shrug:  This is a nice cleanup though and definitely in the intended direction.

jvdydev (approved):
Looks good :smile:

PR #435, opened 2025-03-03
New function `crafted-save-selected-packages`

### LINKED ISSUE #432: [Emacs 30.1] `package-selected-packages` not saved  - including workaround
### Context

One of the core ideas of Crafted Emacs is to think of the Emacs config in three phases:
1. Package phase: add packages to the variable `package-selected-packages`
2. Installation phase: run `package-install-selected-packages`
3. Configuration phase: do the rest.

This has two upsides:
1. Checking whether all the packages you use are actually installed is now very fast, especially if the list of installed packages exists in a file like `custom.el`.
2. Keeping track which packages where installed intentionally by the user and should be left alone by housekeeping functions like `packages-autoremove` is now almost automatic.

### Steps to reproduce

- OS: Linux (Arch)
- Emacs Version: 30.1
- Set Emacs custom file to `custom.el` 
- Set customization variable `crafted-init-auto-save-selected-packages` set to `t` (default)

### What happened?

- After the init process, `package-selected-packages` is set to `nil`, and is written as such to the custom file.
- Consequently, running `package-autoremove` attempts to remove *all* packages.

###  What should have happened?

- After the init process, `package-selected-packages` should reflect the list of intentionally installed packages and should be written as such to the custom file.

### Information about the cause

I tracked the issue to the function `package--save-selected-packages` in Emacs' `package.el`. We add that without parameters to the `after-init-hook` in `crafted-init-config.el`. 

In Emacs <=29, that function reads:
```emacs-lisp
(defun package--save-selected-packages (&optional value)
  "Set and save `package-selected-packages' to VALUE."
  (when value
    (setq package-selected-packages value))
  (if after-init-time
      (customize-save-variable 'package-selected-packages package-selected-packages)
    (add-hook 'after-init-hook #'package--save-selected-packages)))
```
in Emacs 30.1, it reads:
```emacs-lisp
(defun package--save-selected-packages (&optional value)
  "Set and save `package-selected-packages' to VALUE."
  (when (or value after-init-time)
    ;; It is valid to set it to nil, for example when the last package
    ;; is uninstalled.  But it shouldn't be done at init time, to
    ;; avoid overwriting configurations that haven't yet been loaded.
    (setq package-selected-packages (sort value #'string<)))
  (if after-init-time
      (customize-save-variable 'package-selected-packages package-selected-packages)
    (add-hook 'after-init-hook #'package--save-selected-packages)))
```

The culprit seems to be "or" in `(when (or value after-init-time)`, which I believe to be a bug in Emacs. If I read the code correctly, it does the opposite of what the comment says. If it is run after init (as per the hook), it sets the variable to `nil`, no matter what. The "or" should probably be an "and". But I'm not sure if I understand that correctly.

Either way, the effect is that if this function is called without a parameter (as we do it), the variable is set to `nil`.

### Workarounds

There are two ways to resolve this, even while the code remains like this in Emacs.

1. set `crafted-init-auto-save-selected-packages` to `nil`
That way, the function is not run by the `after-init-hook` and the state of the variable after init is preserved.   
2. add the function to `after-init-hook` with a parameter 
For example like this (according to the [Emacs Manual](https://www.gnu.org/software/emacs/manual/html_node/emacs/Hooks.html), this is he way to do it) in `crafted-init-config.el`:

```emacs-lisp
(when crafted-init-auto-save-selected-packages
  (add-hook 'after-init-hook (lambda () (package--save-selected-packages package-selected-packages))))
```

That way, the variable _is_ written to the custom file after init, but reflecting the actual state of the variable after init, no longer overwriting it with `nil`.

This is backward compatible to Emacs 27. I haven't checked further back than that.

### Proposal

I propose the second workaround.

sthesing:
Close #432

### DIFF
--- a/modules/crafted-init-config.el
+++ b/modules/crafted-init-config.el
@@ -167,11 +167,18 @@ we can't reload the file."
   (when custom-file
     (load custom-file :noerror)))
 
+(defun crafted-save-selected-packages ()
+  "Save the list of selected packages.
+
+This saves the state of the variable `package-selected-packages` to the
+custom file after initialization."
+  (package--save-selected-packages package-selected-packages))
+
 ;; Save all customizations to `custom-file', unless the user opted out.
 (when crafted-init-auto-save-customized
   (add-hook 'after-init-hook #'crafted-save-customized))
 (when crafted-init-auto-save-selected-packages
-  (add-hook 'after-init-hook #'package--save-selected-packages))
+  (add-hook 'after-init-hook #'crafted-save-selected-packages))
 
 (provide 'crafted-init-config)
 ;;; crafted-init-config.el ends here


jeffbowman (approved):

jeffbowman:
Looks good to me!

PR #437, opened 2025-03-05
[Emacs 30.1] Add configuration for global-completion-preview

### LINKED ISSUE #431: [Emacs 30.1] Favour built-in `completion-preview-mode` over Corfu?
I only spotted this because it was mentioned in Mickey Peterson's post "[What’s New in Emacs 30.1?](https://www.masteringemacs.org/article/whats-new-in-emacs-301)".

The new `completion-preview-mode`  is a nice completion UI and thus a built-in alternative to Corfu. My tests suggest that it plays well with Cape and the other packages we use in `crafted-completion-*`. 

We should:

- Discuss whether we should favour it over Corfu?
- Implement a basic setup in `crafted-completion-config` if the user does _not_ use Corfu.

jeffbowman:
Adds configuration for global-completion-preview in the crafted-defaults-config module. 

Closes #431

### DIFF
--- a/modules/crafted-completion-config.el
+++ b/modules/crafted-completion-config.el
@@ -30,7 +30,11 @@
     (fido-mode -1)
     (fido-vertical-mode -1)
     (icomplete-mode -1)
-    (icomplete-vertical-mode -1)))
+    (icomplete-vertical-mode -1)
+    (when (version< "30" emacs-version)
+      ;; this mode is only available in Emacs version 30.1 and
+      ;; greater.
+      (global-completion-preview-mode -1))))
 
 
 ;;; Marginalia
--- a/modules/crafted-completion-config.el
+++ b/modules/crafted-completion-config.el
@@ -30,11 +30,7 @@
     (fido-mode -1)
     (fido-vertical-mode -1)
     (icomplete-mode -1)
-    (icomplete-vertical-mode -1)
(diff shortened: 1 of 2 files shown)


sthesing:
Just a quick question: what if a user uses `crafted-completion-config` and has `corfu` installed. Don't they interfere? Shouldn't we turn `global-completion-preview` off in the corfu config?
I won't be able to test until tomorrow. But maybe you already thought of that and tried it?

jeffbowman:
Ah.. good point, I should turn that off, I'll update the PR.

modules/crafted-completion-config.el
```diff
@@ -30,7 +30,11 @@
     (fido-mode -1)
     (fido-vertical-mode -1)
     (icomplete-mode -1)
-    (icomplete-vertical-mode -1)))
+    (icomplete-vertical-mode -1)
+    (when (version< "30" emacs-version)
+      ;; this mode is only available in Emacs version 30.1 and
+      ;; greater.
+      (global-completion-preview-mode -1))))
```
sthesing, line 37: I think this should not go here in the `vertico`-config, but  in the code behind the `(when (require 'corfu nil :noerror)` guard.

sthesing (changes requested):

sthesing (approved):
Thanks!

PR #444, opened 2025-09-29
docs: describe how to disable Crafted Emacs modules

### LINKED ISSUE #443: What is the easiest and proper way to remove packages installed by Crafted Emacs modules?
I use the `crafted-ide` module, which installs the `editorconfig` package.  Because I use Emacs 30 that has `editorconfig` built-in, I decided to remove the installed package.  These were the steps:

1. Add a version check to the `crafted-ide` module
2. Remove the `editorconfig` package using a package manager (in my case elpaca)
3. Remove the package from `package-selected-packages` using Easy Customization

The last step feels superfluous, but is necessary.  Is there a way to do steps 2 and 3 in one?

SergiyKolesnikov:
Resolves #443

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ bookmarks
 network-security.data
 quelpa/
 modules/.dir-locals.el
+.aider*
--- a/docs/getting-started-guide.org
+++ b/docs/getting-started-guide.org
@@ -268,13 +268,13 @@ ls ./modules/*-config.el
 *** Disabling Crafted Emacs Modules
 
 If you no longer want to use a module provided by Crafted Emacs anymore, you can
- disable it by completing the following two steps
+ disable it as follows
 
 1. Stop loading the module in your Emacs initialization file
 2. If the module includes packages from a crafted-*-packages.el file, then uninstall the packages installed by the module
 
-For example, if you previously enabled the =crafted-completions-packages=
-and =crafted-completion-config= modules, find and remove the corresponding
+For example, if you previously enabled the =crafted-completions-packages.el=
+and =crafted-completion-config.el= modules, find and remove the corresponding
 =require= expressions from your Emacs initialization file:
 
 #+begin_src emacs-lisp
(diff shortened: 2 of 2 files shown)


docs/getting-started-guide.org
```diff
@@ -265,6 +265,81 @@ To list all the available configuration modules, you can run:
...
+Next, open the =crafted-completion-packages.el= file and note
+all packages specified in
+=(add-to-list 'package-selected-package ...)= expressions.
+
+Finally, uninstall these packages using either:
+
+- The command ~M-x package-delete~
+- Or the =pakage.el= UI (to open the UI type ~M-x list-packages~)
+
+**** Disabling Only Specific Packages from a Crafted Emacs Module
```
jeffbowman, line 295: This section suggests putting things into the `init.el` file, which is fine, but possibly an "easier" approach is to copy the crafted module to the `custom-modules/` folder and then edit the file from there.  Modules found in the users `custom-modules/` folder are loaded before the ones found in `crafted-emacs`. Thus the `init.el` file does not have to change.  Perhaps we need a generic description of how to customize Crafted Emacs using a method like this and then we can say "see <that chapter>" with some specific commentary on this specific task. 

Does that makes sense?

jeffbowman (changes requested):
Thanks for this addition, but take a look at the comments I left in the code.

docs/getting-started-guide.org
jeffbowman: Correct, the folder is added to the `load-path` only.  The user must load the modules in `custom-modules` with a `require` form (or other ways, but this is the most common way).  

What I mean is this, 

```emacs-lisp
```

does not have to change.  You can copy the `crafted-ide-packages.el` file to your own `custom-modules/` folder, edit there.  Crafted Emacs will find and load the `crafted-ide-packages` from the `custom-modules/` folder as it is earlier in the `load-path`.  Does that help?

docs/getting-started-guide.org
```diff
@@ -271,7 +271,7 @@ If you no longer want to use a module provided by Crafted Emacs anymore, you can
  disable it by completing the following two steps
 
 1. Stop loading the module in your Emacs initialization file
-2. Uninstall the packages installed by the module
+2. If the module includes packages from a crafted-*-packages.el file, then uninstall the packages installed by the module
```
jeffbowman, line 274: This looks good to me.

jeffbowman (approved):

docs/getting-started-guide.org
SergiyKolesnikov: I've updated the section. Please take a look.

.gitignore
```diff
@@ -10,3 +10,4 @@ bookmarks
 network-security.data
 quelpa/
 modules/.dir-locals.el
+.aider*
```
jeffbowman, line 13: what is this?

jeffbowman (changes requested):

.gitignore
SergiyKolesnikov: Session files created by [aider](https://aider.chat/)

PR #445, opened 2025-09-29
feat: add exec-path-from-shell to the IDE module

SergiyKolesnikov:
Many Emacs IDE features depend on external tools, such as, LSP servers, interpreters, compilers, and linters. To ensure Emacs can locate the executables for these tools, both the environment variables and the exec-path variable (see "(Emacs) Shell") must be configured correctly.

### DIFF
--- a/docs/crafted-ide.org
+++ b/docs/crafted-ide.org
@@ -27,7 +27,8 @@ This module includes:
 - LSP client (Eglot)
 - Next-generation syntax parsing (Tree-Sitter)
 - [[https://editorconfig.org][EditorConfig]] support ([[https://github.com/editorconfig/editorconfig-emacs][editorconfig-emacs]]).
-
+- Environment-variables setup
+  
 *** Aggressive Indentation
 
 By default Emacs enables [[https://www.gnu.org/software/emacs/manual/html_node/emacs/Indent-Convenience.html][Electric Indent]], a minor mode that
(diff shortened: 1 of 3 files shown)


jvdydev:
I personally haven't had this issue (granted, I'm not on macOS nor GNOME).

If this only is an issue for select setups (as described), I would lean towards putting it behind a documented customisation flag or as a documented example to avoid loading additional things in environments where this isn't an issue (Windows, more custom Linux setups, ...).

I haven't tested if this also happens on e.g. dmenu/rofi (and related alternatives), I tend to launch Emacs using a WM shortcut or a terminal (both of which seem to inherent the "correct" (expected) environment).

jeffbowman:
I'm kinda swamped at work, I'm looking into this PR locally as I can.  Sorry for the delay.

jeffbowman:
@SergiyKolesnikov , sorry for the delay, finally getting a moment to dig into this more.

Not everyone who uses the IDE module will have a need for those particular variables.  Also, for PYTHONPATH specifically, that may change depending on the virtual environment used and the tools the developer might have installed (ala `anaconda`, `uv-mode`, `pyvenv`, etc) which should probably be preferred over setting this in the environment, but that decision is left to the developer. Thus for this suggestion, installing the package is sufficient as it provides `PATH` and `MANPATH` by default, and the developer can choose to add whatever other variables would be useful to them. 

An alternative would be to only set values if they exist in the environment:

```emacs-lisp
;; ... 
(when (getenv var)
        (add-to-list 'exec-path-from-shell-variables var))
;; ...
```

SergiyKolesnikov:
No worries at all, thanks for digging into this.

> An alternative would be to only set values if they exist in the environment

This is what effectively happens by default. For example, my shell environment does not have `PYTHONPATH` set. Therefore, even though `PYTHONPATH` is added to `exec-path-from-shell-variables`, `(getenv "PYTHONPATH")` will still return `nil`.

The reason for adding those common environment variables during package initialization is to avoid additional overhead of starting a shell multiple times to add them later, as suggested in the [package documentation](https://github.com/purcell/exec-path-from-shell?tab=readme-ov-file#making-exec-path-from-shell-faster).

So, if those common variables are in the user's shell environment, they will be added to the Emacs environment without additional user intervention and startup overhead. If they are not in the shell environment, they will have no effect.

jeffbowman:
Thank you for your contribution and patience!

